PR #31, opened 2023-06-16
Feature/vector output

blooop:
Add support for plotting vector outputs.  Until now bencher has only operated on scalar inputs and outputs and due to the way it is stored in a structured xarray dataset it is difficult to unify the two modes of operation.  This pr creates a dictionary to store the vector results as they are currently incompatible with the scalar datatypes.   This pr only allows for 1D input sweeps and future work is to generalise this across N dimensions (to match the features of scalar inputs and outputs). 

This adds an example of how to use the vector outputs:
image

image

### DIFF
--- a/bencher/bench_vars.py
+++ b/bencher/bench_vars.py
@@ -73,7 +73,7 @@ def param_hash(param_type: Parameterized, hash_value: bool = True, hash_meta: bo
 class ParametrizedSweep(Parameterized):
     """Parent class for all Sweep types that need a custom hash"""
 
-    def hash_custom(self) -> str:
+    def hash_persistent(self) -> str:
         """A hash function that avoids the PYTHONHASHSEED 'feature' which returns a different hash value each time the program is run"""
         return param_hash(self, True, False)
 
@@ -81,7 +81,7 @@ class ParametrizedSweep(Parameterized):
 class ParametrizedOutput(Parameterized):
     """Parent class for all Output types that need a custom hash"""
 
-    def hash_custom(self) -> str:
+    def hash_persistent(self) -> str:
         """A hash function that avoids the PYTHONHASHSEED 'feature' which returns a different hash value each time the program is run"""
         return param_hash(self, True, False)
 
@@ -161,7 +161,7 @@ class BoolSweep(Boolean):
         """Generate a string representation of the sampling procedure"""
         return f"sampling {self.name} from: [True,False]"
 
-    def hash_custom(self) -> str:
+    def hash_persistent(self) -> str:
         """A hash function that avoids the PYTHONHASHSEED 'feature' which returns a different hash value each time the program is run"""
         return hash_extra_vars(self)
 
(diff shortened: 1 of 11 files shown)


bencher/bench_vars.py
```diff
@@ -12,12 +12,9 @@
 from sys import byteorder
 
 
-def hash_cust(var: any):
+def hash_cust(var: any) -> str:
```
mfinean: Naming of hash_cust and hash_custom is confusing/very close

bencher/bench_vars.py
```diff
@@ -76,14 +73,16 @@ def param_hash(param_type: Parameterized, hash_value: bool = True, hash_meta: bo
...
+    def hash_custom(self) -> str:
+        """A hash function that avoids the PYTHONHASHSEED 'feature' which returns a different hash value each time the program is run"""
         return param_hash(self, True, False)
 
 
 class ParametrizedOutput(Parameterized):
     """Parent class for all Output types that need a custom hash"""
 
-    def hash_custom(self) -> int:
+    def hash_custom(self) -> str:
```
mfinean: ParametrizedOutput and ParametrizedSweep have the same class definition - are we able to remove duplication here?

bencher/bench_vars.py
```diff
@@ -504,3 +511,80 @@ def index_names(self) -> List[str]:
...
+        index_name: str,
+        index_units: str,
+        default=ResultSeries(),
+        units="ul",
+        indices: List[float] = None,
+        instantiate=True,
+        **params,
+    ):
+        param.Parameter.__init__(self, default=default, instantiate=instantiate, **params)
+        # self.
```
mfinean: Remove commented lines

bencher/bencher.py
```diff
@@ -383,13 +387,14 @@ def setup_dataset(
                 result_data = np.empty(dims_cfg.dims_size)
                 result_data.fill(np.nan)
                 data_vars[rv.name] = (dims_cfg.dims_name, result_data)
-            else:
+            elif type(rv) == ResultVec:
                 for i in range(rv.size):
                     result_data = np.empty(dims_cfg.dims_size)
```
mfinean: np.empty followed by a fill can be done together with  
`np.full( *SHAPE*, *VALUE*)`

mfinean (changes requested):
Main comments are with respect to class duplication and updated naming for the 'hash_custom'-type functions to make it clearer what these are/when they're used

bencher/bench_vars.py
blooop: I was planning on doing this in a separate pr cos it involves a large find/replace operation.

#32

bencher/bench_vars.py
blooop: hash_cust -> hash_sha1
has_custom -> hash_persistent

mfinean (approved):

PR #35, opened 2023-06-21
Feature/plot sig

blooop:
Apologies for the large PR.

This is part 1 of the plotting refactor.  The idea is to replace the large if/else tree in plt_cfg.py with a system for automatically selecting what plot to display by providing a library of available plotters and a filtering system where the plot describes what type of data it can plot.

This enables multiple plots for the same data and greater flexibility around defining custom plots or other representations of the data.

The new code is left inactive by default.  Future prs will add more plot providers but should be a lot smaller in scope as the filtering machinery has been added in this pr. 

image


bencher/bench_cfg.py
```diff
@@ -103,8 +103,11 @@ def as_dict(self, include_params: List[str] = None, exclude_params: List[str] =
 
         return output
 
+    # def as_sns_args(self) -> dict:
```
mfinean: Should this commented code be removed?

bencher/plotting/plot_collection.py
```diff
@@ -0,0 +1,76 @@
...
+    def __init__(self) -> None:
+        """A PlotCollection has a dictionary of function_name:function_pointer which are all the available types of plot. The plotters variable contains an ordered dictionary of the active plotting functions"""
+        self.plotter_providers = {}
+        self.plotters = {}
+
+    def add_plotter_source(self, plotter_class_instance: PlotProvider) -> None:
+        """Given a class definition that contains several plotting methods, get a list of all the methods of the class and add them to the dictionary of available plotting functions
+
+        Args:
+            plotter_class_instance (PlotProvider): A class that contains plotting methods
```
mfinean, line 22: Can also add to docstring

Raises:
    ValueError: Explanation here.

bencher/plotting/plot_collection.py
```diff
@@ -0,0 +1,76 @@
...
+    ) -> pn.pane.panel:
+        """This method returns a single plot based on 1 result variable and a set of input variables.  It dedeuces the correct plot type by passing it to several configuration functions that operate on the number of inputs
+
+        Args:
+            bench_cfg (BenchCfg): A config of the input vars
+            rv (ParametrizedSweep): a config of the result variable
+            plt_cnt_cfg (PltCntCfg): A config of how many input types there are"""
+
+        tabs = pn.Tabs()
+        for p in self.plotters.values():
```
mfinean: p used as tmp variable for nested loops looks confusing

bencher/plotting/plot_filter.py
```diff
@@ -0,0 +1,108 @@
...
+    def matches(self, val: int) -> bool:
+        """Checks that a value is within the variable range.  lower_bound and upper_bound are inclusive (lower_bound<=val<=upper_bound )
+
+        Args:
+            val (int): A positive integer representing a number of items
+
+        Returns:
+            bool: True if the items is within the range, False otherwise.
+        """
+        if val >= 0:
```
mfinean: This might look clearer if we have         

`if val < 0:
raise ValueError("val must be >= 0")

...
return lower_match and upper_match
`

bencher/plotting/plot_library.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @staticmethod
+    def all() -> PlotCollection:
+        """Display the data in every possible way. All plots are included
+
+        Returns:
+            PlotCollection: All possible plots
+        """
+        plt_col = PlotLibrary.setup_sources()
+        for pt in [p for p in PlotTypes]:  # iterate through all plot enums
```
mfinean: This looks like it can be simplified to

`for pt in PlotTypes:`

bencher/plotting/plots/tables.py
```diff
@@ -0,0 +1,27 @@
+from typing import List
+import panel as pn
+from bencher.plotting.plot_types import PlotTypes
+from bencher.plotting.plot_filter import PlotInput
+
+
+class Tables:
+    """A class to display the result data in tabular form"""
+
+    def dataframe_flat(self, pl_in: PlotInput) -> List[pn.panel]:
```
mfinean, line 10: What's the reason for returning as a list (with 1 entry in all)?

test/test_plot_collection.py
```diff
@@ -0,0 +1,78 @@
...
+        assert "plot_1" in plot_coll.plotters
+
+    # Tests that removing a plot successfully removes the plot from the list of active plotting functions
+    def test_remove_plot_success(self):
+        plot_coll = PlotCollection()
+
+        plot_coll.add_plotter_source(TestPlotProvider())
+        plot_coll.add("plot_1")
+        plot_coll.remove("plot_1")
+        assert len(plot_coll.plotters) == 0
```
mfinean, line 42: plot_coll could probably do with a method like get_num_plots()?

blooop: I don't need to know the number of plots in general. this is just a test to assert that add and remove works properly

bencher/plotting/plots/tables.py
blooop: The plot providers can return between 0 and N plots.  By always returning a list I can combine the results of all the plot providers without needed checks for None return values.

bencher/plotting/plot_collection.py
blooop: fixed:

for plt_fn in self.plotters.values():
            plots = plt_fn(PlotInput(bench_cfg, rv, plt_cnt_cfg))
            for plt_instance in plots:
                tabs.append(plt_instance)

bencher/plotting/plot_filter.py
mfinean: if val >= 0 is no longer needed since the complement is caught in the previous if

mfinean (changes requested):

mfinean (approved):

PR #37, opened 2023-06-23
Feature/plot collection add list

blooop:
add add_list function and tests

### DIFF
--- a/test/test_plot_collection.py
+++ b/test/test_plot_collection.py
@@ -82,7 +82,8 @@ class TestPlotCollection(unittest.TestCase):
 
         plot_coll.add_plotter_source(TestPlotProvider())
         with self.assertRaises(ValueError):
-            plot_coll.add_list("plot_1")  # try adding a variable that is not a list
+            # try adding a variable that is not a list
+            plot_coll.add_list("plot_1")
 
     def test_add_list_empty(self):
         plot_coll = PlotCollection()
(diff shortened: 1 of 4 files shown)


test/test_plot_collection.py
```diff
@@ -70,9 +70,21 @@ def test_remove_nonexistent_plot(self):
             plot_coll.remove("plot_3")
         except KeyError:
             pass
-        assert len(plot_coll.plotters) == 1
```
mfinean: What is this testing for with the assertion removed?

test/test_plot_collection.py
```diff
@@ -70,9 +70,21 @@ def test_remove_nonexistent_plot(self):
...
 
     def test_add_uninitialised_class(self):
         plot_coll = PlotCollection()
         with self.assertRaises(ValueError):
             plot_coll.add_plotter_source(TestPlotProvider)
+
+    def test_add_list_scalar(self):
```
mfinean: Why is this _scalar? It's unclear what this means since it adds "plot_1" (a string?)

test/test_plot_collection.py
blooop: nice catch, that was an accidental removal

test/test_plot_collection.py
blooop: Adding a scalar variable (instead of vector variable).   I have updated the test name and added a comment.

mfinean (approved):

PR #38, opened 2023-06-24
set up the plot library in the initialisation of bencher

blooop:
set up the plot library at the beginning of bencher which is then shared by all of the plot_sweep function calls

### DIFF
--- a/bencher/bencher.py
+++ b/bencher/bencher.py
@@ -22,8 +22,7 @@ from bencher.bench_vars import (
 from bencher.plt_cfg import BenchPlotter
 from bencher.bench_cfg import BenchCfg, BenchRunCfg, DimsCfg
 from bencher.bench_plot_server import BenchPlotServer
-from bencher.plotting.plot_library import PlotLibrary
-from bencher.plotting.plot_types import PlotTypes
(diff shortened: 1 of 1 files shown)


bencher/bencher.py
```diff
@@ -98,6 +98,7 @@ def __init__(
         bench_name: str = None,
         worker: Callable = None,
         worker_input_cfg: ParametrizedSweep = None,
+        plot_lib: PlotCollection = None,
```
mfinean, line 101: Missing entry in the docstring to say what this is

bencher/bencher.py
```diff
@@ -207,7 +209,7 @@ def plot_sweep(
         )
 
         bench_cfg.param.update(run_cfg.param.values())
-        bench_cfg.plot_lib = plot_lib
+        bench_cfg.plot_lib = plot_lib if plot_lib is not None else self.plot_lib
```
mfinean, line 213: Similarly in this plot_sweep method, plot_lib needs adding to the docstring

mfinean (changes requested):

mfinean (approved):

PR #42, opened 2023-06-28
Feature/pcolormesh

blooop:
add imshow() plot method to the plot library


test/test_plot_library.py
```diff
@@ -10,7 +10,8 @@ class TestPlotLibrary(unittest.TestCase):
     def test_default_plot_collection(self) -> None:
         default_plots = PlotLibrary.default()
         self.assertIn(PlotTypes.swarmplot, default_plots.plotters)
-        self.assertEqual(len(default_plots.plotters), 1)
+        self.assertIn(PlotTypes.imshow, default_plots.plotters)
+        self.assertEqual(len(default_plots.plotters), 2)
```
mfinean: Worth a comment for these saying why default_plots has two plotters

mfinean (changes requested):

test/test_plot_library.py
blooop: I am just going to delete the test entirely.   It doesn't really serve any purpose and increases the amount of work when adding new plots.

PR #44, opened 2023-06-28
Feature/refactor plot return types

blooop:
I can't find the comment, but in a previous pr the return types of the empty list and list of plots was questioned.  I have addressed that comment here and simplified the return types of the plots.  They either return a single panel or none.  The logic of the gather_plots() has been updated to reflect the changes.

The import order was auto updated by ruff.


bencher/plotting/plot_collection.py
```diff
@@ -86,14 +87,24 @@ def gather_plots(
...
+        """
 
         tabs = pn.Accordion()
         for plt_fn in self.plotters.values():
             plots = plt_fn(PlotInput(bench_cfg, rv, plt_cnt_cfg))
-            for plt_instance in plots:
-                logging.info(f"plotting: {plt_instance.name}")
-                tabs.append(plt_instance)
+            if plots is not None:
+                if type(plots) != list:
```
mfinean, line 99: What's the motivation for this change? Is this trying to catch the case where plots has a length of 1?

blooop: before all plots would be contained in a list. Now the plots are either a single plot or potentially a list of plots.  This just sets the plots to the most compatible format (a list of plots) for the rest of the processing.

mfinean (approved):

PR #49, opened 2023-07-03
Feature/scatter2d

blooop:
add 2d scatter plots to plot library


bencher/plotting/plots/scatterplot.py
```diff
@@ -0,0 +1,46 @@
...
+
+from bencher.plotting.plot_filter import PlotFilter, PlotInput, VarRange
+from bencher.plotting.plot_types import PlotTypes
+from bencher.plotting_functions import wrap_long_time_labels
+
+
+class Scatter:
+    """A class to display the result data as a scatter plot"""
+
+    def scatter2D_sns(self, pl_in: PlotInput) -> pn.panel:
```
mfinean: Also returns None

mfinean (approved):

bencher/plotting/plots/scatterplot.py
blooop: fixed

PR #253, opened 2023-11-21
Feature/dimension grid

blooop:


blooop:
## Recommend closing — every file this touches is gone

Opened 2023-11-21. Checked against `main` today:

| Path in this PR | Status |
|---|---|
| `bencher/dimension_grid.py` | never existed on main |
| `bencher/plotting/plots/volume.py` | gone — volume plotting now lives in `bencher/results/volume_result.py` |
| `index.html` | gone from the repo root |

The plotting layer has been through the plugin-registry rewrite and the plot-filter
work since, so there is no rebase target left. If the dimension-grid idea is still
wanted it belongs as a fresh proposal against the current results/plugin architecture —
`plans/architecture/A2-plot-selection-redesign.md` is the natural home.

PR #543, opened 2025-02-19
Feature/extract image

blooop:
## Summary by Sourcery

This pull request introduces a new feature to extract a frame from a video at a specific time. It also adds an example of how to use this feature.

New Features:
- Added a function to extract a frame from a video at a given time and save it as a PNG image.

Enhancements:
- The example now extracts a frame from the generated video and saves it as an image.


blooop:
closing as the docs seem to be using a cached environment with the wrong version of sqlite

PR #688, opened 2025-12-26
Split bench_cfg.py into focused configuration modules

mfinean:
## Summary
- Split the large bench_cfg.py into 10 focused modules under bench_cfg/
- Group related parameters: cache, execution, display, visualization, time, server
- Use composition instead of deep inheritance for BenchRunCfg
- Maintain full backward compatibility via dynamic delegation

## Changes
- New `bencher/bench_cfg/` package with focused config classes
- Original `bench_cfg.py` becomes a thin re-export layer
- No changes to public API or parameter names/defaults

## Summary by Sourcery

Modularize benchmark configuration into focused submodules while preserving the existing public API via a thin compatibility layer.

New Features:
- Introduce dedicated configuration classes for caching, execution, display, visualization, time, and plot server settings under the new bencher.bench_cfg package.
- Add a composed BenchRunCfg that groups the sub-configuration objects while still supporting flat parameter access and command-line construction.
- Expose BenchCfg, BenchRunCfg, DimsCfg, and the new sub-config classes through the bencher.bench_cfg package for both grouped and flat configuration usage.

Enhancements:
- Refactor the previous monolithic bench_cfg.py into smaller, concern-focused modules, improving structure and maintainability without changing parameter names or defaults.

Tests:
- Add unit tests covering default values of the new sub-config classes, BenchRunCfg composition and delegation behavior, BenchCfg inheritance, and backward-compatible import paths.


bencher/bench_cfg/cache_cfg.py
```diff
@@ -0,0 +1,60 @@
...
+        "aggregate of all the results at the end of the benchmark. This cache lets you stop a "
+        "benchmark halfway through and continue. However, beware that depending on how you "
+        "change code in the objective function, the cache could provide values that are not "
+        "correct.",
+    )
+
+    only_hash_tag: bool = param.Boolean(
+        False,
+        doc="By default when checking if a sample has been calculated before it includes the "
+        "hash of the greater benchmarking context. This is safer because it means that data "
```
blooop, line 36: Weird multiple line strings?

bencher/bench_cfg/execution_cfg.py
```diff
@@ -0,0 +1,33 @@
...
+class ExecutionCfg(param.Parameterized):
+    """Execution settings: sampling, parallelization, and run modes."""
+
+    repeats: int = param.Integer(1, doc="The number of times to sample the inputs")
+
+    level: int = param.Integer(
+        default=0,
+        bounds=[0, 12],
+        doc="The level parameter is a method of defining the number samples to sweep over in a "
+        "variable agnostic way, i.e you don't need to specify the number of samples for each "
```
blooop, line 19: More

blooop:
I like the look of this in principle, but less sure when it comes to the implementation. Ie before the code was "bad" because it was monolithic, but was relatively simple because it was all flat and made all the hashing logic easy to verify. Now it's "better" and more modular, but seems there there is more room for error and hashing bugs. There is also more logic now to go wrong. Maybe this is covered in the tests? I'm not sure if this will be easier to maintain.

bencher/bench_cfg/cache_cfg.py
mfinean: I thknk the multi-line strings were from ruff auto-formatting - can revert to single lines if preferred

mfinean:
> I like the look of this in principle, but less sure when it comes to the implementation. Ie before the code was "bad" because it was monolithic, but was relatively simple because it was all flat and made all the hashing logic easy to verify. Now it's "better" and more modular, but seems there there is more room for error and hashing bugs. There is also more logic now to go wrong. Maybe this is covered in the tests? I'm not sure if this will be easier to maintain.

Your call :) The hashing logic is unchanged - hash_persistent still accesses self.repeats, self.over_time etc. which now delegate to sub-configs but return identical values. The tests verify delegation works and hash behavior is preserved

blooop:
Can't test on computer, but does bench_cfg.xxx still auto complete as before?  If it does then happy to try this.   If it doesn't would an inheritance approach work?

blooop:
Now I have been able to check out on my computer and see in more detail, looks good.

PR #689, opened 2025-12-26
Fix flaky hypothesis tests in test_job.py

mfinean:
## Summary

- Increased hypothesis test deadlines from 500-2000ms to 5000ms for all tests in test_job.py
- The multiprocessing executor has significant startup overhead that was causing intermittent deadline failures

## Summary by Sourcery

Introduce a structured benchmark configuration system with composable sub-config classes and update tests and configuration access accordingly.

Bug Fixes:
- Relax Hypothesis deadlines in job tests to avoid flaky failures caused by multiprocessing startup overhead.

Enhancements:
- Add BenchRunCfg as a composable run-configuration that delegates parameters to sub-config objects while preserving flat, backward-compatible access.
- Add BenchCfg as a higher-level benchmark configuration encapsulating variables, metadata, hashing, and sweep description utilities.
- Provide dedicated cache, execution, display, visualization, time, server, and dims configuration classes to better organize benchmark settings.
- Refactor the bench_cfg package into a proper subpackage with explicit exports and improve LaTeX/markdown description helpers for sweeps and results.

Tests:
- Add comprehensive tests for sub-config defaults, BenchRunCfg composition/delegation, BenchCfg behavior, and backward-compatible imports.

### DIFF
--- a/test/test_job.py
+++ b/test/test_job.py
@@ -19,7 +19,7 @@ class CachedParamExample(bch.CachedParams):
 
 
 class TestJob(unittest.TestCase):
-    @settings(deadline=2000)  # Increased from 500ms to 2000ms
+    @settings(deadline=5000)  # Increased deadline for multiprocessing startup overhead
     @given(st.sampled_from([bch.Executors.SERIAL, bch.Executors.MULTIPROCESSING]))
     def test_basic(self, executor):
         cp = CachedParamExample()  # clears cache by default
@@ -49,7 +49,7 @@ class TestJob(unittest.TestCase):
         res1cp3 = jc3.call(var1=1).result()
         self.assertNotEqual(res1["result"], res1cp3["result"])
 
-    @settings(deadline=500)
+    @settings(deadline=5000)  # Increased deadline for multiprocessing startup overhead
     @given(st.sampled_from([bch.Executors.SERIAL, bch.Executors.MULTIPROCESSING]))
     def test_overwrite(self, executor):
         cp = CachedParamExample()  # clears cache by default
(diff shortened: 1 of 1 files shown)


blooop:
Are these new flaky test or old tests?  The existing tests are not flaky so if this is increasing the deadlines on old tests this indicates a problem with the implementation.  (Sorry on phone so difficult to see the trace of the test heritage)

mfinean:
> Are these new flaky test or old tests? The existing tests are not flaky so if this is increasing the deadlines on old tests this indicates a problem with the implementation. (Sorry on phone so difficult to see the trace of the test heritage)

I'm fairly sure they're existing flaky tests.

blooop:
Flaky on ci or your machine?  I have ironed out all flaky tests a long time ago

blooop:
My meaning being, if they have become flaky now, it indicates a performance regression

mfinean:
Ah, interesting. Right, so CI runs on Linux (platforms = ["linux-64"] in pyproject.toml)
I'm currently working on macOS which uses spawn for multiprocessing (reimports all modules per subprocess). Linux uses fork which is much faster. So the ~2000ms overhead is macOS multiprocessing startup, not a code regression.

Interesting though. Could be worth increasing the deadlines anyway on the tests then and/or testing on different platforms?

blooop:
Ah ok, that doesn't sound like a performance regression after all then, and yes could add Mac to test matrix

mfinean:
Just updated - was based off the other PR. This was meant to be a standalone test fix

blooop:
Yeah, but given it was based on the previous pr which changed cache/hash behaviour I wondered if those changes were causing performance regressions.  It sounds like that is not the case though and is related to  platform differences.

blooop (approved):

PR #750, opened 2026-03-06
Claude/refactor holoviews backend v exu e

blooop:


blooop:
Closing in favor of #754, which is a cleaned-up version of this same refactoring with test updates included.

PR #772, opened 2026-03-16
Fix YamlSweep stale cache causing identical hashes

blooop:
## Summary
- Removed unreliable `lru_cache` from `YamlSweep._read_yaml` that used `(path, mtime_ns, size)` as cache key
- When YAML content changed but file size stayed the same (e.g., `param: 1` → `param: 2`) and writes happened faster than filesystem mtime resolution, stale cached data was returned
- This caused `hash_persistent()` to produce identical hashes for different YAML configs, making `test_yaml_sweep_hash_includes_value` fail when run with the full test suite

## Test plan
- [x] `test_yaml_sweep_hash_includes_value` now passes when run with the full suite (`pixi run pytest test/test_yaml_sweep.py`)
- [x] Full test suite passes (555 passed, 5 skipped)

## Summary by Sourcery

Fix YAML input loading to avoid stale cached content and ensure configuration changes are reflected in hashing.

Bug Fixes:
- Prevent YAML sweep hashes from remaining identical when YAML file contents change but filesystem metadata does not update accordingly.

Enhancements:
- Simplify YAML loading by removing an unreliable LRU cache layer and consolidating reading logic into a single helper.

### DIFF
--- a/bencher/variables/inputs.py
+++ b/bencher/variables/inputs.py
@@ -390,7 +390,7 @@ class YamlSweep(SweepSelector):
         )
 
     @staticmethod
-    def _load_yaml(path: Path) -> Mapping[str, Any]:
+    def _load_yaml(path: Path) -> Any:
         with path.open("r", encoding="utf-8") as stream:
             data = yaml.safe_load(stream)
         return data if data is not None else {}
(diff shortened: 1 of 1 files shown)


blooop:
Addressed Sourcery's review feedback:

1. **Type hint mismatch**: Fixed in 1d68a87 — changed `_load_yaml` return type from `Mapping[str, Any]` to `Any` since `yaml.safe_load` can return non-mapping types. The caller already validates with `isinstance(entries, Mapping)`.

2. **Performance**: No action needed — `_load_yaml` is only called once during `__init__`, and YAML config files are small, so the removed cache has negligible impact.

PR #853, opened 2026-03-27
fix: generate optimisation reports for all benchmarks

blooop:
## Summary
- Three call sites used `bench.results[-1].to_optuna_plots()`, which only produced optimisation plots for the **last** `BenchResult`
- When a `Bench` has multiple results (from multiple `plot_sweep()` calls), earlier benchmarks were silently skipped
- Changed all three sites to loop over all results so every benchmark gets its optimisation report

## Files changed
- `bencher/run.py` — `_with_optimise` wrapper
- `bencher/bencher.py` — `Bench.optimize()` with `plot=True`
- `bencher/example/meta/generate_examples.py` — doc generation

## Test plan
- [ ] `pixi run ci` passes (format, lint, tests with coverage)
- [ ] Run a benchmark with multiple `plot_sweep()` calls and `optimise > 0` — verify all benchmarks get optimisation plots

## Summary by Sourcery

Generate optimisation reports for all benchmark results instead of only the last one when running optimisations.

Bug Fixes:
- Ensure optimisation reports are generated for every BenchResult in Bench.optimize instead of just the most recent result.
- Fix example generation so documentation benchmarks with multiple runs produce optimisation reports for all results.
- Update the optimisation wrapper to attach reports for all available benchmark results when optimisation completes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.73.1] - 2026-03-27
+
+### Fixed
+- Generate optimisation reports for all benchmarks, not just the last (#852)
+- Include time event in over_time panel labels and default to last (#849)
+- Wrap long description strings in generated examples to fit 100-char line limit (#851)
+- Override RTD theme CSS to wrap long lines in docs code blocks (#851)
+
 ## [1.73.0] - 2026-03-26
 
 ### Added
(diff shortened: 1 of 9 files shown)


blooop:
Thanks for the review @sourcery-ai! Addressing the suggestions:

1. **Duplicate reports**: `optimize()` is called once per `run()` invocation, so there's no risk of duplicate reports being appended.

2. **Extract a helper**: The loop appears in only two places (`bencher.py:optimize()` and `run.py:_with_optimise`) with different error-handling contexts — extracting a shared helper would be premature abstraction at this point.

3. **Per-result try/catch in `_with_optimise`**: The `try`/`except` is already inside the `for res in bench.results` loop (see `run.py:142-149`), so a failure in one result's `to_optuna_plots()` won't prevent subsequent results from being processed.

PR #918, opened 2026-04-19
Add regression_percent_floor for dual-band adaptive detection

blooop:
## Summary

- Adds an optional `regression_percent_floor` knob to `BenchRunCfg` that layers a percent-change acceptance band on top of the `adaptive` method's MAD band.
- With the floor set, a regression fires only when BOTH tests reject — MAD-based z-score AND percent change exceed their thresholds. `None` preserves current behavior.
- Primary use case: low-repeat variables where the MAD-derived noise floor is very tight, so a modest real move trips the step test even though the absolute percent change is small in practice.

## Design

- `detect_adaptive` takes a new `percent_floor` parameter. The step gate uses `_safe_change_percent(curr_mean, baseline)`; the drift gate uses the projected end-of-history change so a trend whose total projected drift is below the floor is allowed.
- `RegressionResult` now carries `percent_band_lower` / `percent_band_upper`; the existing `band_lower` / `band_upper` still describe the MAD band.
- The PNG and holoviews diagnostic renderers draw both bands when both are present (MAD band in the verdict colour, percent band in purple), and relabel the MAD band as "MAD band" when the percent band is also shown.

## Test plan

- [x] `pixi run pytest test/test_regression.py` — 72 passed
- [x] `pixi run lint` — 10.00/10 pylint, no new ruff/ty findings
- [x] New unit tests cover: percent floor suppresses MAD-only fire on tight history; percent floor preserves real regressions; `percent_floor=None` is identical to the pre-change behavior; percent gate is direction-aware; integration test end-to-end via `detect_regressions` with a low-repeat dataset.

## Summary by Sourcery

Introduce an optional dual-band percent-change floor to the adaptive regression detector and propagate it through configuration, detection, and visualization.

New Features:
- Add a regression_percent_floor option to BenchRunCfg that sets a minimum percent change required to flag regressions for the adaptive method.
- Extend adaptive regression detection to combine MAD-based tests with an optional percent-change gate, only firing when both thresholds are exceeded.
- Augment RegressionResult and regression plots to carry and render a second percent-based acceptance band alongside the MAD band.

Tests:
- Add unit and integration tests covering percent_floor behavior, configuration wiring, direction awareness, and preservation of existing behavior when unset.

### DIFF
--- a/bencher/bench_cfg.py
+++ b/bencher/bench_cfg.py
@@ -375,6 +375,18 @@ class BenchRunCfg(BenchPlotSrvCfg):
         "Useful for failing CI pipelines on benchmark regressions.",
     )
 
+    regression_percent_floor: float = param.Number(
+        default=None,
+        allow_None=True,
+        doc="Optional minimum percent change required to flag a regression, applied as a "
+        "second acceptance band on top of the adaptive method's MAD band. A regression "
+        "only fires when BOTH the MAD-based test AND the percent change exceed their "
+        "thresholds. Useful for suppressing noise-floor false positives on metrics with "
+        "few repeats: set e.g. 40.0 to require at least a 40%% change regardless of how "
+        "many MAD-sigma it is. If None (default), only the MAD band gates. "
+        "Only affects 'adaptive' method.",
+    )
+
     def __init__(self, **params: Any) -> None:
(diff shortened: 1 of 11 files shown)


blooop:
Superseded by #922, which combines this work with the regression plot renderer rewrite.

PR #704, opened 2026-01-24
Revert PR #688: Return to monolithic bench_cfg.py with improved organization

blooop:
## Summary
- Reverts PR #688's modular configuration structure back to monolithic `bench_cfg.py`
- Adds clear section comments to improve organization within the single file
- Preserves all functionality and backward compatibility

## Why Revert?

After critical analysis, the monolithic approach is objectively better for bencher because:

1. **Simpler architecture** - No complex delegation magic with `__getattr__`/`__setattr__`
2. **Better CLI discovery** - All parameters available from single `bench_cfg.` autocomplete
3. **Natural param library integration** - The param library is designed for flat parameter classes
4. **Reduced complexity** - 680 lines in one file is manageable with proper organization
5. **No real isolation benefit** - Config parameters rarely change independently

### 1. Reverts the modular structure
- Removes 11 separate config files 
- Restores single `bench_cfg.py` (680 lines)
- Removes complex delegation system
- Eliminates two-tiered access patterns

### 2. Improves organization with section comments
```python
# ==================== EXECUTION PARAMETERS ====================
# ==================== CACHE PARAMETERS ====================  
# ==================== DISPLAY PARAMETERS ====================
# ==================== VISUALIZATION PARAMETERS ====================
# ==================== TIME & HISTORY PARAMETERS ====================
```

### 3. Groups related parameters logically
- Parameters are now grouped by functionality
- Easy to find and modify related settings
- Clear separation of concerns within single file

## Test Plan
- [x] All existing tests pass
- [x] `pixi run ci` completes successfully
- [x] Examples run correctly (tested with `example_simple_float.py`)
- [x] Parameter access works via `cfg.parameter_name`
- [x] CLI autocomplete discovery functional
- [x] No breaking changes to public API

🤖 Generated with [Claude Code](https://claude.ai/code)

## Summary by Sourcery

Revert the modular benchmark configuration back into a single consolidated bench_cfg module while preserving behavior and improving in-file organization.

New Features:
- Provide a unified BenchRunCfg/BenchCfg/DimsCfg configuration module that centralizes execution, caching, display, visualization, and time/history parameters in one file.

Enhancements:
- Reorganize configuration parameters into clearly documented sections for execution, caching, display, visualization, and time/history concerns to improve discoverability and maintenance.
- Simplify the configuration architecture by removing the multi-file delegation-based bench_cfg package and using direct param-based classes instead.

Tests:
- Remove the bench configuration unit test suite that targeted the previous modular configuration structure.

PR #923, opened 2026-04-20
Add plan to re-split bench_cfg.py with a clean API break

blooop:
Designs a non-backward-compatible re-enablement of PR #688: nested
sub-configs (CacheCfg, ExecutionCfg, DisplayCfg, VisualizationCfg,
TimeCfg, RegressionCfg, ServerCfg) composed into BenchRunCfg via
param.ClassSelector slots, with no delegation magic and no flat-access
shims. Includes parameter-to-group mapping, migration steps, and
acceptance criteria.

## Summary by Sourcery

Documentation:
- Add BENCH_CFG_SPLIT_PLAN.md describing the target bench configuration module layout, parameter grouping, migration steps, and acceptance criteria for the future refactor.


blooop:
## Superseded — this has already been absorbed into A5

Flagging rather than closing, since the call is yours.

`plans/architecture/A5-config-surface-reduction.md` is in `main` today and explicitly
takes this on: the plans index describes A5 as *"Four orthogonal config objects (spec /
run / analysis / report); one home per knob; **absorbs PR #923 as its Phase 1**;
per-field disposition for all 73 `BenchCfg` fields"*. A5 also draws the same lesson
from the #688 → #704 revert that this document opens with — kill the flat layer rather
than maintain dual access — and then generalizes it past `BenchRunCfg` to the whole
73-field surface.

Two further reasons not to land the file as-is:

1. **Wrong location and naming.** It adds `BENCH_CFG_SPLIT_PLAN.md` at the repo root,
   but the convention is `plans/NN-kebab-name.md` (or `plans/architecture/AN-*.md`).
   Plan 03 has since been through and tidied root-level clutter.
2. **Its own inventories are stale.** The document warns that its parameter counts were
   generated on 2026-07-23 and must be regenerated before execution. Several phases of
   plan 23 have reworked config-adjacent code since, and `bench_cfg.py` is currently the
   single largest source of type-check debt (14 of the 125 outstanding Tier-B
   diagnostics), so the inventory would need redoing before anyone could act on it.

**Recommendation:** close, and if anything here isn't already reflected in A5 — the
concrete rename table is the most valuable part — fold that into A5's Phase 1 section
instead. Happy to do that merge-forward if you want it.

PR #926, opened 2026-04-22
Make regression report reflect per-method gate semantics

blooop:
## Summary

- The regression report formatter stamped ``+NN% change (baseline=…, threshold=T)`` on every row, regardless of which detector produced it. For ``detect_absolute`` that rendered ``+nan%`` and labelled the hard limit as ``Baseline``; for ``detect_delta`` it showed a percent even though the gate is on the raw delta.
- ``RegressionResult.summary()`` and ``RegressionReport.to_markdown()`` now dispatch per method so each row describes its real gate:
  - **percentage**: threshold shown as ``±T%``
  - **adaptive**: threshold shown as ``Tσ`` (change still in percent)
  - **delta**: Change column shows raw ``Δ``, threshold as ``±T``
  - **absolute**: Change and Baseline cells are em-dash (no historical baseline), threshold carries direction-aware inequality (``≤`` / ``≥``); summary phrased as ``current=X vs ceiling|floor=Y``
- Detection logic, ``RegressionResult`` shape, and callers are untouched — formatting-only change.

### Example rendering

```
Regressions detected in 4 variable(s):
  response_time_ms: current=27 vs ceiling=25 (method=absolute)
  throughput_rps: current=80 vs floor=100 (method=absolute)
  latency_pct: +20.00% change (baseline=100, current=120, method=percentage, threshold=±5%)
  latency_delta: Δ=+8 (baseline=100, current=108, method=delta, threshold=±5)
```

| Variable | Change | Baseline | Current | Method | Threshold |
|----------|-------:|----------:|--------:|--------|----------:|
| response_time_ms | — | — | 27 | absolute | ≤ 25 |
| throughput_rps | — | — | 80 | absolute | ≥ 100 |
| latency_pct | +20.0% | 100 | 120 | percentage | ±5% |
| latency_delta | +8 | 100 | 108 | delta | ±5 |

## Test plan

- [x] ``pixi run pytest test/test_regression.py`` (90/90 passing, 10 in ``TestRegressionReport``)
- [x] ``pixi run ruff format`` / ``pixi run ruff check`` clean on both files
- [x] ``prek run --files bencher/regression.py test/test_regression.py`` clean
- New tests cover: percentage exact format (pinned), adaptive threshold has ``σ`` units, delta shows raw ``Δ`` and hides percent, absolute minimize/maximize for both ``summary()`` and ``to_markdown()``, absolute non-regressed renders without NaN leak.

## Summary by Sourcery

Adjust regression report rendering to reflect the semantics of each detection method rather than always showing percentage changes.

New Features:
- Add per-method summary formatting for regression results, including specialized rendering for percentage, adaptive, delta, and absolute methods.
- Add per-method markdown table row formatting so the Change and Threshold columns reflect the correct gate units and semantics.

Bug Fixes:
- Prevent absolute-method regression summaries and markdown output from showing NaN percentage changes or mislabeling limits as baselines.

Tests:
- Add regression report tests covering per-method summary and markdown output, including percentage, adaptive (σ thresholds), delta (raw Δ), and absolute methods for both regressed and non-regressed cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,21 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.90.0] - 2026-04-22
+
+### Added
+- Regression report is now auto-embedded as a Markdown panel at the top of `to_auto_plots()` whenever `regression_report.has_regressions` is true. Previously only the per-variable overlay plots were injected, so absolute-method fires (which have no history/overlay) were silent in the report.
+
+### Changed
+- Regression report rendering (`RegressionReport.summary()` and `to_markdown()`) now dispatches per method so each row describes its actual gate:
+  - `percentage`: threshold shown as `±T%`.
+  - `adaptive`: threshold shown as `Tσ` (change remains in percent).
+  - `delta`: Change column shows the raw Δ (not percent, since the gate is in absolute units); threshold rendered as `±T`.
+  - `absolute`: Change and Baseline cells rendered as em-dash (no historical baseline); Threshold cell carries the direction-aware inequality (`≤ L` for `OptDir.minimize`, `≥ L` for `OptDir.maximize`). Summary line phrased as `current=X vs ceiling|floor=Y`.
+
+### Fixed
+- `RegressionResult.summary()` / `RegressionReport.to_markdown()` no longer render `+nan%` or mislabel the hard limit as `Baseline` for `regression_method="absolute"` results.
+
 ## [1.89.0] - 2026-04-21
 
 ### Added
--- a/bencher/regression.py
+++ b/bencher/regression.py
@@ -152,71 +152,105 @@ class RegressionReport:
         report.prepend_to_result(bench_res, md)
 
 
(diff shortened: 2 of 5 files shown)


blooop:
Thanks for the review! Pushed cea3daaa addressing all three comments:

1. **Consolidation** — added a frozen `_MethodCells` dataclass in `_method_cells(r)` that captures the per-method rendering (change / baseline / threshold / summary lead). Both `_format_summary_line` and `_format_markdown_row` are now thin consumers of that bundle, so method semantics live in one place.
2. **Comment mismatch (absolute branch)** — the original comment said "baseline_value is the limit" while the code read from `threshold`. Both fields hold the limit for the absolute detector, but `threshold` is the semantically clearer choice; I kept the code reading `threshold` and dropped the stray `baseline_value` reference from the comment.
3. **Adaptive summary pin + negative-change coverage** — adaptive now asserts the full expected summary string (matching the percentage/delta tests). Added three negative-change tests (`test_percentage_output_negative`, `test_adaptive_output_negative`, `test_delta_output_negative`) pinning the signed formatting (`-50.00% change`, `Δ=-8`) for summary and markdown.

On the typing suggestion for `direction`: `RegressionResult.direction` is declared and stored as `str` (each `detect_*` writes `direction=direction.value`), so accepting `str` matches the field. Typing it as `OptDir` would require reshaping `RegressionResult`, which is outside the scope of this formatting-only change.

PR #940, opened 2026-05-15
Add fullscreen + open-in-new-tab controls to rerun viewer

blooop:
## Summary
- Embedded rerun recordings now show two small overlay controls at top-center: ⛶ for fullscreen and ↗ to open the same chromeless viewer in a new tab. Useful when an over_time sweep renders multiple recordings side-by-side and you want to expand one without losing the others.
- New tab uses a real \`<a target="_blank">\` (with the href populated on click from the iframe's current src), not \`window.open\` — popup blockers leave anchor clicks alone but silently drop \`window.open\` in some browsers.
- Controls read their target from the iframe DOM, so the existing URL rewriting in \`inline_rrd_iframes()\` for sidecar/portable saves still only needs to touch the iframe. Verified the post-processing regex still matches after wrapping.
- Bumps version to \`1.100.0\` and adds a CHANGELOG entry.

## Test plan
- [x] \`pixi run lint\` — clean
- [x] Manual: live Panel server rendering an over_time sweep — fullscreen button enters/exits fullscreen, open-in-new-tab opens the chromeless viewer in a new tab.
- [ ] Verify saved (sidecar) and portable (inline-base64) reports still rewrite iframe URLs correctly with the wrapper present.

## Summary by Sourcery

Add overlay controls to embedded rerun viewer iframes for fullscreen and opening the viewer in a new tab, and bump the package version.

New Features:
- Provide a wrapper around rerun viewer iframes that adds top-centered overlay controls for fullscreen and opening the viewer in a new browser tab.

Enhancements:
- Ensure embedded and portable rerun viewer panes use the new iframe wrapper while preserving existing URL rewriting behavior.

Build:
- Bump the project version from 1.99.0 to 1.100.0.

Documentation:
- Document the new overlay controls and behavior in the changelog for version 1.100.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.100.0] - 2026-05-15
+
+### Added
+- Overlay controls on each embedded rerun recording: a fullscreen button (⛶) that calls `iframe.requestFullscreen()` and an open-in-new-tab link (↗) that opens the same chromeless viewer in a new browser tab. Useful when comparing multiple recordings side-by-side and you want to expand one. Controls are positioned top-center to avoid the viewer's own corner UI.
+
 ## [1.99.0] - 2026-05-15
 
 ### Changed
--- a/bencher/utils_rrd.py
+++ b/bencher/utils_rrd.py
@@ -127,7 +127,7 @@ def rrd_file_to_pane(  # pragma: no cover
         f' frameborder="0" allowfullscreen></iframe>'
     )
     return pn.pane.HTML(
-        _wrap_viewer_controls(iframe, width, height),
+        _wrap_viewer_controls(iframe, viewer_url, width, height),
         width=width,
         height=height,
     )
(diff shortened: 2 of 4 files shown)


blooop:
Addressed the legit bug Sourcery flagged in comment 1:

- **Middle-click / Cmd-click / context-menu "open in new tab"** previously got `about:blank` because the anchor's `href` was only set in the `onclick` handler. Fixed by writing `viewer_url` into the `href` at construction time. `inline_rrd_iframes()` was extended to rewrite both occurrences (iframe `src` + anchor `href`) in a single pass over the saved HTML, memoized per `rrd_path` so sidecar/inline-viewer files aren't written twice.

Skipping the other suggestions:

- **`document.fullscreenEnabled` guard** — the iframe is rendered inside the same document (already has `allowfullscreen`), and the API is supported in every browser version Panel targets. A guarded "do nothing if unsupported" call is worse UX than a console error on an exotic browser.
- **Percentage-based wrapper sizing** — the pane is fixed-pixel because `result_var.width`/`height` are pixel-valued. Switching the wrapper to `100%` doesn't unlock responsive sizing while the pane itself is fixed.
- **Null-checks on `closest().querySelector()`** — DOM is constructed by us a few characters earlier in the same template; the only way the lookup is null is if the wrapping HTML has been mangled, in which case silently no-op'ing is worse than an error.
- **Comment 2 (constants + multiline template refactor)** — cosmetic, scope creep on a small helper.

PR #941, opened 2026-05-18
Add bencher.ci module for CI regression gating

blooop:
## Summary

Extracts generic CI integration utilities into a new `bencher.ci` module. These functions bridge the gap between bencher's regression detection engine and CI workflow plumbing (GitHub Actions, Jenkins, etc.):

- **`write_performance_summary()`** — serialize a `RegressionReport` to a pipe-delimited text file that CI can parse, with optional per-metric filtering and custom thresholds
- **`render_regression_plots()`** — render diagnostic PNGs for regressed metrics (for embedding in PR comments)
- **`warn_on_regressions()`** — one-call convenience wrapper: write summary + render plots + emit pytest warnings
- **`parse_performance_summary()`** — parse the summary file back into structured dicts
- **`generate_regression_comment()`** — produce a GitHub-flavored Markdown PR comment with status icons, table, plot links, and next-steps guidance

### Motivation

Projects using bencher for benchmark regression detection end up reimplementing the same CI glue: writing structured summaries, generating PR comments with regression tables, rendering plot PNGs. This was ~150 lines of fragile bash in CI workflows plus ~100 lines of Python in downstream projects. Moving it upstream means:

1. One tested implementation instead of per-project copies
2. New detection methods (adaptive, delta, absolute) automatically render correctly in comments
3. `method_cells()` stays the single source of truth for how each method is displayed

### Design choices

- **No new dependencies** — uses only stdlib + existing bencher imports
- **`metrics_filter` is optional** — without it, every result in the report is written. With it, only matching metrics appear and per-metric thresholds override the report-level ones
- **`bench_name` prefix** — allows qualified names like `"bench_planning/planning_time"` so identically-named metrics from different benchmarks are distinguishable
- **`append=True` default** — multiple benchmark tests write to the same summary file during a CI run
- **Comment generation is pure Python** — replaces `bash` + `awk` + `sed` in CI YAML, testable and maintainable

### Example usage

```python
from bencher.ci import warn_on_regressions, generate_regression_comment

# In a pytest benchmark test:
warn_on_regressions(
    bench.results[-1].regression_report,
    summary_path=Path("reports/performance_summary.txt"),
    plot_dir=Path("reports/regression_plots"),
    bench_name="bench_robot_planning",
    metrics_filter={"bench_robot_planning/planning_time": 20.0},
)

# In a CI step (or Python script called from CI):
comment_md = generate_regression_comment(
    "reports/performance_summary.txt",
    report_url="https://reports.example.com/latest/index.html",
    plot_url_prefix="https://reports.example.com/latest/regression_plots",
)
```

## Test plan

- [x] 26 new tests covering all functions (`test/test_ci.py`)
- [x] Existing 103 regression tests still pass
- [ ] Verify linting passes (`pixi run ci`)

## Summary by Sourcery

Introduce a bencher.ci module providing reusable CI utilities for benchmark regression gating, along with comprehensive tests.

New Features:
- Add write_performance_summary to serialize regression reports into CI-friendly pipe-delimited summaries with optional filtering and thresholds.
- Add render_regression_plots to generate PNG diagnostics for regressed metrics, with optional metric filtering and bench name prefixing.
- Add warn_on_regressions as a convenience wrapper to emit pytest warnings and optionally produce summaries and plots from a regression report.
- Add parse_performance_summary to read performance summary files back into structured dictionaries for further processing.
- Add generate_regression_comment to build GitHub-flavored Markdown PR comments summarizing benchmark regressions and linking reports and plots.

Tests:
- Add dedicated test suite in test/test_ci.py covering summary writing, plot rendering, regression warnings, summary parsing, and comment generation behavior.

### DIFF
--- a/bencher/__init__.py
+++ b/bencher/__init__.py
@@ -124,6 +124,13 @@ from .regression import (
     MethodCells,
     method_cells,
 )
+from .ci import (
+    write_performance_summary,
+    render_regression_plots,
+    warn_on_regressions,
+    parse_performance_summary,
+    generate_regression_comment,
+)
 from .perf_tracker import PerfTracker, PerfReport
 from .git_info import git_time_event
 from .plotting.plot_filter import VarRange, PlotFilter
--- a/bencher/ci.py
+++ b/bencher/ci.py
@@ -1,0 +1,337 @@
+"""CI integration utilities for benchmark regression gating.
+
+Provides functions to:
+- Write structured performance summaries parseable by CI workflows
+- Generate GitHub-flavored Markdown for PR comments
+- Render regression plot PNGs for regressed metrics
+- Emit pytest warnings on detected regressions
+
+All functions operate on :class:`~bencher.regression.RegressionReport` and
+:class:`~bencher.regression.RegressionResult` — no framework-specific
+dependencies beyond bencher itself.
+"""
+
+from __future__ import annotations
+
+import logging
+import warnings
+from pathlib import Path
+
+from bencher.regression import RegressionReport, RegressionResult, method_cells
+
+logger = logging.getLogger(__name__)
+
+_SUMMARY_HEADER = "variable|regressed|change_percent|baseline|current|threshold|direction|method"
+
+
+def write_performance_summary(
+    report: RegressionReport,
+    path: Path | str,
+    *,
+    metrics_filter: dict[str, float] | None = None,
+    bench_name: str = "",
+    append: bool = True,
+) -> list[str]:
+    """Write regression results to a pipe-delimited text file for CI parsing.
+
+    Each line encodes one variable comparison::
+
+        variable|regressed|change_percent|baseline|current|threshold|direction|method
+
+    Args:
+        report: Regression report from a benchmark run.
+        path: Output file path. Created (with parents) if it does not exist.
+        metrics_filter: Optional allowlist mapping ``"bench_name/variable"``
+            (or just ``"variable"`` when *bench_name* is empty) to a
+            per-metric regression threshold (percentage).  When provided,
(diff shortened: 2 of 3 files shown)


blooop:
## Superseded in approach by plan 10 — worth reconciling before landing

`plans/10-regression-policy-and-verdict-export.md` covers this ground and lands on the
opposite side of the central design question, so merging as-is would commit us to the
interface plan 10 exists to retire.

This PR's contract is a **pipe-delimited text file** plus a `parse_performance_summary()`
to read it back. Plan 10's stated goal is *"a canonical, schema-versioned JSON contract
plus a one-call aggregate (`report.has_gate_regressions`) so a merge gate needs zero
parsing"*. A bespoke text format with a hand-written parser on the far side is exactly
the out-of-band coupling being designed away — and it has no schema version, so any
later field change silently breaks every consumer.

Plan 10 also introduces first-class **severity** (`gate` / `notify` / `info`) carried on
the result variable itself. `warn_on_regressions()` here flattens that distinction to a
single warn-or-not, so a notify-only regression can't feed a Slack message without also
looking like a gate failure.

The genuinely useful piece that plan 10 does *not* cover is `render_regression_plots()` —
diagnostic PNGs for embedding in PR comments. Note that regression rendering is plan 08
Task 1's territory (it extracts the rendering half into `regression_rendering.py`), so
that function wants to land there rather than in a new `bencher/ci.py`.

**Recommendation:** close, and fold the plot-rendering helper into plan 08 Task 1 and the
CI-gating surface into plan 10's export layer. The `bencher.ci` namespace itself is a
good idea — it just wants plan 10's JSON contract underneath it.

PR #942, opened 2026-05-19
fix: Show all repeat recordings for media results

blooop:
## Problem

With `repeats > 1`, the bencher report only showed one recording for `ResultRerun`, `ResultVideo`, and `ResultImage` results — even though all recordings were correctly created on disk and stored in the xarray dataset.

**Root cause**: Two interacting issues in `_to_panes_da`:

1. **Pane recursion peeled the repeat dimension** — `_to_panes_da` treated `repeat` like any other dimension and iterated over it via `_iter_pane_slices`, calling `ds_to_container` for each individual repeat. `ds_to_container` → `zero_dim_da_to_val` then extracted a single scalar value, so each repeat rendered independently but only the first was visible in the final layout.

2. **`zero_dim_da_to_val` picked `values[0]`** — Even when the repeat dimension survived (e.g. via SQUEEZE in the rerun path at `rerun_result.py:86`), the function always returned the first element.

The `over_time` dimension already had special handling for this exact scenario (`_pane_over_time_grid` / `_pane_over_time_slider`), but the `repeat` dimension had no equivalent.

## Fix

Three changes in `bench_result_base.py`:

1. **Exclude `repeat` from pane recursion for media results** — mirrors the existing `over_time` exclusion. Prevents `_to_panes_da` from peeling the repeat dimension before the grid handler can process it.

2. **Route to `_pane_repeat_grid`** — after the over_time check, checks for a surviving repeat dimension on media results and delegates to the new grid method.

3. **`_pane_repeat_grid` method** — iterates each repeat value and renders a labelled `pn.Row` of panels. Handles `ResultRerun` (via `rrd_file_to_pane`), `ResultVideo` (via `pn.pane.Video`), and `ResultImage` (via `pn.pane.PNG`) with a fallback for unknown types.

The existing `zero_dim_da_to_val` fix (taking last instead of first) acts as a safety net for any path that doesn't hit the grid.

## Futureproofing

- **`over_time` + `repeat` combo**: If both dimensions survive, `over_time` is handled first (returns early). The repeat grid only fires when over_time is absent or has been consumed.
- **New media result types**: The grid has explicit branches for Rerun/Video/Image with a Markdown fallback.
- **`repeats=1`**: SQUEEZE drops the repeat dimension entirely — no behavioral change.
- **Numeric results unaffected**: The `isinstance` guard ensures only media types bypass pane recursion.

## Test plan

- [ ] Add standalone example with `ResultRerun` + `repeats > 1` (no external dependencies)
- [ ] Add unit test verifying `_pane_repeat_grid` is called when repeat dim > 1 for media results
- [ ] Verify upstream `example_rerun_sweep.py` still works (no repeat dimension)
- [x] Validated with real benchmarks using `repeats=2` and `repeats=5` — all recordings visible

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,24 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.101.0] - 2026-05-19
+
+### Fixed
+- Media results (`ResultRerun`, `ResultVideo`, `ResultImage`) with `repeats > 1` now render every repeat instead of silently dropping all but one. Previously the repeat dimension was peeled by generic pane recursion and `zero_dim_da_to_val` quietly returned `values[0]`, leaving N-1 recordings on disk but invisible in the report.
+
+### Changed
+- `zero_dim_da_to_val` is now strict: it raises `ValueError` when handed a DataArray with a multi-valued dimension, instead of silently collapsing it. Callers must peel multi-valued dims (via recursion, `isel`, or `_pane_media_grid`) first.
+- Extracted shared `_render_media_cell` helper used by both `_pane_media_grid` and `_pane_over_time_slider` so new media types only need to be wired up in one place.
+- Generalised the over-time-specific grid into `_pane_media_grid`, which iterates the cartesian product of every surviving multi-valued dim (`over_time`, `repeat`, future custom dims) and lays out cells as a `pn.Row` (1-D) or `pn.GridBox` with `ncols` (N-D). `_pane_over_time_grid` is kept as a thin shim.
+
+### Added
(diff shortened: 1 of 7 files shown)


blooop:
Thanks @sourcery-ai — the refactor in c1fce89b (and follow-up e41823c5) covers your three concerns:

1. **`zero_dim_da_to_val` silently selecting `-1` along every dim** — now strict: it raises `ValueError` when handed a multi-valued dim. Callers (`_pane_over_time_slider`, `_pane_media_grid`, the holoview tap stream) peel surviving dims explicitly via `isel` before calling. No more silent semantic changes for non-media callers.
2. **`_pane_repeat_grid` duplicating `_pane_over_time_grid`** — both paths now funnel through a single `_pane_media_grid` helper, with a shared `_render_media_cell` doing the `Rerun/Video/Image` dispatch. `_pane_over_time_grid` is kept as a one-line shim for backwards compatibility.
3. **`zero_dim_da_to_val` policy documentation** — the new docstring spells out the strict contract.

The media-type `isinstance` tuple is now only used in one place (the dispatch in `_to_panes_da`), so a dedicated `_is_media_result` helper would be one wrapper around one call site — not extracting that for now.

PR #943, opened 2026-05-31
Add collect/render split for out-of-process report rendering

mfinean:
## Problem

Building a report (holoviews/panel/bokeh) allocates large object graphs backed by C-extension wrappers (param, pandas, bokeh). When CPython's cyclic GC traverses those wrappers while **foreign live C-extension state** exists in the same interpreter — e.g. ROS 2 `rclpy`/DDS in a robotics benchmark — the process can **segfault** (exit -11).

Today the only entry point is `plot_sweep`, which always builds the report in-process immediately after collecting samples. So consumers running alongside such C-extensions have no way to keep holoviews out of the live process; `gc.disable()` doesn't help (it only suppresses *automatic* collection, and the report build is what creates the pressure).

## Change

Make the two phases separable so rendering can happen in a clean process that never imported the foreign extension:

- **`plot_sweep(auto_plot=...)`** — new parameter, default `True` (behaviour unchanged). When `False`, the sweep runs and regression detection is computed, but **no** holoviews/panel objects are constructed. Applied via `run_cfg` so it survives the `run_cfg → bench_cfg` param merge in `run_sweep` (`auto_plot` lives on `BenchRunCfg`, which `BenchCfg` inherits — setting it only on the `BenchCfg` constructor would be clobbered by the merge).
- **`Bench.collect(...)`** — thin wrapper for `plot_sweep(auto_plot=False)`; returns a fully-populated, picklable `BenchResult` (dataset + `regression_report`).
- **`bencher.render_report()` / `save_result()` / `load_result()`** (new `bencher/render` module) — persist a collected result and render the HTML report from it, optionally in a separate process via `python -m bencher.render <result> <out_dir>`.

The result dataset and `RegressionReport` are already serialisable (`cache_results` pickles the whole `BenchResult`; the render module strips/restores the non-picklable `object_index` the same way), so the render step needs nothing live.

## Usage

```python
# Process holding rclpy/DDS — builds no plots:
res = bench.collect(input_vars=[...], result_vars=[...])
bencher.save_result(res, "result.pkl")

# Clean process (e.g. spawned subprocess):
#   python -m bencher.render result.pkl out_dir
bencher.render_report("result.pkl", "out_dir")
```

## Validation

- New `test/test_render.py` (11 tests): `collect()` appends no report tabs and constructs far fewer holoviews/bokeh objects than a render; `auto_plot=False` via `plot_sweep` matches; `collect()` rejects an explicit `auto_plot`; save/load round-trip; `render_report` from object and from path; the `python -m bencher.render` CLI (incl. bad-args/missing-file exit codes).
- End-to-end: collect (0 plot objects) → save → **fresh subprocess** render → valid HTML.
- Existing suite unaffected (`test_bencher.py`/`test_bench_report.py`/`test_bench_result_base.py`: 91 passed, 1 skipped). ruff + pylint clean; prek hooks pass.
- New `bencher/example/example_collect_render.py`.

## Notes

Additive and backwards-compatible — the default `auto_plot=True` path is unchanged. Motivated by recurring GPU-benchmark segfaults in a ROS 2 consumer (kinisi_benchmarks); this is the upstream half of the fix so any consumer running alongside foreign C-extensions benefits.

## Summary by Sourcery

Introduce a collect/render split for benchmark reports to allow out-of-process HTML rendering and avoid in-process plotting issues with foreign C extensions.

New Features:
- Add an auto_plot flag to plot_sweep to optionally skip building plots while still computing benchmark results.
- Introduce Bench.collect as a non-plotting sweep entrypoint that returns a fully populated BenchResult for later rendering.
- Add a bencher.render module with save_result, load_result, render_report, and a python -m bencher.render CLI for out-of-process report generation.
- Provide an example script demonstrating the collect/render workflow for benchmarks.

Tests:
- Add test_render.py to validate collect behavior, serialization round trips, render_report functionality, and the bencher.render CLI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,3 +30,23 @@ jobs:
         uses: codecov/codecov-action@v6
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+
+  # Re-run the suite with the collect/render split forced on, so every report
+  # build goes through save_result -> load_result -> render. Catches any divergence
+  # between in-process and split rendering (unpicklable result types, render paths
+  # that relied on live state). See bencher/render.py and test/test_render.py.
+  ci-split:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+      - uses: prefix-dev/setup-pixi@v0.9.5
+        with:
+          cache: true
+          frozen: true
+          environments: py313
+      - name: Forced collect/render split
+        run: |
+          pixi update
+          pixi run -e py313 generate-examples
+          pixi run -e py313 test-split
--- a/.gitignore
+++ b/.gitignore
@@ -196,6 +196,7 @@ docs/autoapi/*
 
 # Claude Code local settings (personal, not shared)
 .claude/settings.local.json
+.claude/worktrees/
 _tabs/*
 
 # Auto-generated docs (HTML reports, RST files) — examples are tracked
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -86,6 +86,14 @@ Bencher is a benchmarking framework built around these core concepts:
 - Coverage reporting with coverage.py
 - Examples serve as integration tests
 - Meta-generated examples in `bencher/example/meta/`
+- **Collect/render split coverage** (`bencher/render.py`): three layers guard the
+  out-of-process render pipeline against divergence from the normal `plot_sweep` path —
+  (A) parity tests in `test/test_render.py` assert `collect()` computes the same dataset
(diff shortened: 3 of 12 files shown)


mfinean:
Added `test_save_result_preserves_object_index` in a9d8674d. I tweaked the suggestion slightly: rather than asserting the example benchmark produces a non-empty `object_index` (it doesn't — no ResultReference/video result vars), the test injects a sentinel list and asserts `save_result` leaves it unchanged (same identity + contents). That makes the invariant deterministic regardless of result-var types.

blooop:
Can you set this as the default so everything uses the split method? That will mean the existing test suite will run over the new method.

PR #952, opened 2026-06-03
Flip ResultFloat/ResultVec/ResultBool default from 0 to NaN

blooop:
## What

Flip the default value of `ResultFloat`, `ResultVec`, **and `ResultBool`** from `0` to `NaN`.

## Why

A `0`/`False` default makes an **unrecorded** sample — a run that aborts before measuring, or a result var the worker never sets — indistinguishable from a real `0`/`False` measurement, dragging nan-aware regression/aggregation means toward zero. The storage layer (`result_collector.setup_dataset`) already initialises result arrays with `np.nan`; the `0` param default was the value that *overwrote* that "unwritten" sentinel.

After this change, unrecorded samples stay `NaN` and are dropped by the existing nan-aware reductions (`skipna=True` xarray reductions, `_clean_1d` in `regression.py`, `np.nanmean`/`nansum` in the agg registry, `df.dropna()` in the Optuna path).

## Scope / decisions

- **`ResultBool` is now included.** An earlier draft kept it at `0`/`False`, justified by the binomial-std calc treating bool means as proportions over a fixed repeat count. That rationale is now obsolete: the binomial SE was fixed on `main` to divide `p*(1-p)` by the per-cell count of valid (non-NaN) repeats, so missing repeats no longer understate it. The semantics are **missing ≠ failure**: an unrecorded repeat is dropped from the success proportion rather than counted as `False`. A worker that wants a crash/abort to count as a failure must explicitly record `False` on its failure path.
- Callers wanting the old zero/False-fill behaviour opt out with `default=0`.
- **`CACHE_VERSION` is deliberately NOT bumped.** The result-var `default` is not part of `BenchCfg.hash_persistent()`, so existing benchmark and `over_time` history caches are preserved (no forced global wipe). The trade-off: the new `NaN` default only applies to cache *misses*; already-cached cells keep whatever sentinel they were stored with, so a benchmark with missing samples may transiently hold a mix of `0` (old) and `NaN` (new) until those cells are recomputed. For rolling `over_time` history this self-heals as old samples age out of the window.

## Verification

`pixi run ci` green (1474 passed, 5 skipped); format/lint clean (pylint 10.00/10). No `allow_nan=False` JSON sinks exist; the report-export contract coerces non-finite floats to `null` via `_finite_or_none`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.105.0] - 2026-06-11
+
+### Changed
+- The default value for `ResultFloat`, `ResultVec`, and `ResultBool` is now `NaN` instead of `0`. An *unrecorded* sample — a run that aborts before measuring, or a result var the worker never sets — is now treated as missing and dropped by the nan-aware regression/aggregation reductions, instead of masquerading as a real `0`/`False` measurement and dragging means toward zero. This matches the storage layer, which already initialises result arrays with `NaN`. Callers who want unrecorded samples to read as `0` can opt out with `default=0`.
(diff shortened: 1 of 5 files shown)


blooop:
Addressed Sourcery review (commit 91eae7d9):

- **Add ResultBool default tests** ✅ — added `test_bool_default_stays_false_not_nan` asserting `ResultBool().default == 0`, falsy, and not-NaN, locking in the intentionally-different boolean default and guarding against an accidental flip to NaN.
- **Centralise NaN in the JSON-serialisation layer** — not applicable: bencher does not serialise results via `json`. Collected results are pickled (`save_result`/`load_result`) and plots are serialised by bokeh (which encodes NaN as `null`). Both paths are covered by the serialization round-trip tests added in this PR (`TestNanDefaultSerialization`), so a NaN default is already handled end-to-end.

PR #953, opened 2026-06-03
Centralise result-variable missing-value representation

blooop:
> ~~Stacked on #952~~ — #952 has merged and `main` has been merged back into this branch; the PR now targets `main` directly with only the centralisation refactor.

## What

Introduce a single source of truth for how a *missing*/unrecorded result entry is represented, and route the storage + consumer code through it.

## Why

"Missing" was represented by three dtype-specific sentinels:

| Result type | dtype | sentinel |
|---|---|---|
| `ResultFloat`/`Bool`/`Vec` (+ numeric) | `float` | `NaN` |
| `ResultReference`/`DataSet` | `int` | `-1` |
| `ResultPath`/`Video`/`Image`/`String`/`Container`/`Rerun` | `object` | `"NAN"` |

…but that mapping was **duplicated** across `setup_dataset` (initial fill) and `_sentinel_for_result_var` (over_time aging), and consumers hardcoded the check per call site (`== "NAN"` for file panes, `np.isnan`/`skipna` for numerics). Adding a new result type meant editing several `isinstance` ladders.

## Change

New helpers in `bencher/variables/results.py`:
- `result_missing_fill(rv) -> (fill_value, numpy_dtype)`
- `result_is_missing(rv, value) -> bool`

- `ResultCollector.setup_dataset` and `_sentinel_for_result_var` now build from `result_missing_fill` — the two `isinstance` ladders collapse into one polymorphic call.
- The hardcoded `== "NAN"` file checks in `bench_result_base.py` become `result_is_missing(result_var, filepath)`.

## Not a behaviour change

The stored fill values and dtypes are **identical** to before — so the on-disk cache format, the golden `BenchCfg` hash, and every reduction are unchanged. `pixi run ci` green (1445 passed, 5 skipped — same as base), format/lint/ty/pylint clean.

## Follow-up enabled (not in this PR)

Because the object-type sentinel is now defined and checked in one place, switching it from the `"NAN"` string to a literal `None` becomes a one-line, consumer-invisible change (would need its own `CACHE_VERSION` bump since it changes the stored value).

## Summary by Sourcery

Centralise the representation of missing result-variable entries and route dataset setup and consumers through shared helpers.

Enhancements:
- Introduce shared helpers to define and detect missing values for result variables across all result types.
- Refactor dataset initialisation and over-time aging to derive fill values and dtypes from the central missing-value helpers instead of per-call-site type ladders.
- Update file-based result rendering to use the shared missing-value predicate rather than hardcoded string sentinels.

### DIFF
--- a/bencher/results/bench_result_base.py
+++ b/bencher/results/bench_result_base.py
@@ -906,13 +906,8 @@ class BenchResultBase:
         )
         html_list = []
         for idx, _t in enumerate(time_vals):
-            ds_t = dataset.isel(over_time=idx)
-            value = self.zero_dim_da_to_val(ds_t[result_var.name])
-            if result_is_missing(result_var, value):
-                html_list.append(_NO_DATA_HTML)
-                continue
-            filepath = str(value)
-            if not os.path.isfile(filepath):
+            filepath = self._over_time_filepath(dataset, result_var, idx)
+            if filepath is None:
                 html_list.append(_NO_DATA_HTML)
                 continue
             if is_rerun:
@@ -953,6 +948,21 @@ class BenchResultBase:
 
         return pn.Column(pn.pane.Bokeh(div), pn.pane.Bokeh(bokeh_slider))
 
+    def _over_time_filepath(self, dataset: xr.Dataset, result_var, idx: int) -> str | None:
+        """Resolve the on-disk filepath for a file-backed result var at an over_time index.
+
+        Returns None when the entry is missing/unrecorded (per ``result_is_missing``)
+        or when the stored path does not point at an existing file.
+        """
+        ds_t = dataset.isel(over_time=idx)
+        value = self.zero_dim_da_to_val(ds_t[result_var.name])
+        if result_is_missing(result_var, value):
+            return None
+        filepath = str(value)
+        if not os.path.isfile(filepath):
+            return None
+        return filepath
+
     def _pane_over_time_grid(
         self,
         dataset: xr.Dataset,
@@ -972,12 +982,8 @@ class BenchResultBase:
 
         items = []
         for idx, label in enumerate(labels):
-            ds_t = dataset.isel(over_time=idx)
-            value = self.zero_dim_da_to_val(ds_t[result_var.name])
-            if result_is_missing(result_var, value):
-                continue
-            filepath = str(value)
-            if not os.path.isfile(filepath):
+            filepath = self._over_time_filepath(dataset, result_var, idx)
+            if filepath is None:
                 continue
             pane = rrd_file_to_pane(filepath, width=result_var.width, height=result_var.height)
             items.append(pn.Column(pn.pane.Markdown(f"**{label}**"), pane))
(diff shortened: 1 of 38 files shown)


blooop:
Addressed Sourcery review (commit 10e2dc84):

- **ResultVec dtype consistency** ✅ — the `ResultVec` branch in `setup_dataset` now uses `fill, dtype = result_missing_fill(rv)` and `np.full(..., fill, dtype=dtype)`, matching the single-column path.
- **Avoid `str()` before the missing check** ✅ — both file-pane consumers now call `result_is_missing(result_var, value)` on the raw dataset value and only `str()` it after confirming it's not missing, so missingness isn't tied to the string form of the sentinel.
- **Document `None`-as-missing for numerics** ✅ — expanded the `result_is_missing` docstring to state that NaN-backed types treat both NaN and `None` as missing (intentional), while `-1`/`"NAN"` types use exact equality.
- **Tie `DATA_VAR_RESULT_TYPES` to `SCALAR_RESULT_TYPES`** — already built as `SCALAR_RESULT_TYPES + _REFERENCE_MISSING_TYPES + _OBJECT_MISSING_TYPES` with an explanatory comment, so a new scalar type added to `SCALAR_RESULT_TYPES` is picked up automatically.

CI green on the stacked branch: 1446 passed, 5 skipped.

blooop:
Applied review fixes and merged `main` (10e2dc84..1039917a):

- **Merge commit 27bb1016** resolves the stale-stack conflicts in favour of what actually merged to main: `ResultBool` default stays **NaN** (1.105.0 behaviour, including the `_validate_bounds` NaN carve-out), `CACHE_VERSION` stays at `"4"` (no bump needed — this refactor stores identical fill values), and the stale 1.103.0 changelog entry / `test_bool_default_stays_false_not_nan` are dropped.
- **1039917a** tightens `result_is_missing`: non-numeric values are no longer float-coerced for NaN-backed types (the *string* `"nan"` is real data, not the missing sentinel); detection uses `numbers.Real`.
- Adds `test/test_result_missing.py` pinning the (fill, dtype) mapping per result type, `DATA_VAR_RESULT_TYPES` membership, `result_is_missing` semantics per sentinel family, the `result_collector` sentinel wrapper, and a fill → typed-array → detect round trip.
- Version 1.107.0 + changelog entry for the refactor.

Local `pixi run ci`: 1490 passed, 5 skipped, format/lint/pylint clean. One unrelated failure (`test_cartesian_pil_renderer.py::test_timeline` golden image hash) — the files it depends on are byte-identical to main, so it's a devcontainer font-rendering difference, not a regression.

bencher/results/bench_result_base.py
```diff
@@ -968,8 +973,11 @@ def _pane_over_time_grid(
         items = []
         for idx, label in enumerate(labels):
             ds_t = dataset.isel(over_time=idx)
-            filepath = str(self.zero_dim_da_to_val(ds_t[result_var.name]))
-            if filepath == "NAN" or not os.path.isfile(filepath):
+            value = self.zero_dim_da_to_val(ds_t[result_var.name])
```
blooop: this block seems duplicated above

blooop: Good catch — extracted the duplicated filepath-resolution block (zero_dim_da_to_val → result_is_missing → os.path.isfile) into a single `_over_time_filepath(dataset, result_var, idx)` helper, now used by both `_pane_over_time_slider` and `_pane_over_time_grid`. Pushed in 2ea75d72.

PR #958, opened 2026-06-10
Allow NaN default/value on ResultBool

blooop:
## What

`ResultBool` rejected `NaN` as a default or value, even though `NaN` is the documented "missing"/unrecorded sentinel for result variables (see `ResultFloat.__init__`).

## Why

`ResultBool` locks its bounds to `[0, 1]`. param validates a Parameter's default against its bounds whenever a **subclass overrides** it, and validates **every value assignment**. So two things broke:

- `ResultBool(default=float("nan"))` raised `must be at most 1, not nan` as soon as a subclass overrode the inherited Parameter (a base-class declaration happens to skip this validation, so the failure only showed up on override — surprising and hard to diagnose).
- Assigning `float("nan")` at runtime to mark a sample missing raised the same.

Net effect: the NaN "missing" sentinel that already works for `ResultFloat`/`ResultVec` was unusable for `ResultBool`. (Concretely, this bit a downstream benchmark that hoisted a `ResultBool` success metric onto a shared base class and overrode it in one subclass.)

## How

Override `ResultBool._validate_bounds` to treat `NaN` as in-bounds, then delegate to the normal bounds check. NaN becomes a valid default and value; genuinely out-of-range values (e.g. `2.0`) are still rejected.

## Tests

Added to `test/test_result_nan_default.py`:
- `ResultBool` default can be NaN; default stays `0` for backward compat.
- A subclass overriding an inherited NaN-default `ResultBool` no longer raises (the regression).
- Assigning `float("nan")` at runtime is accepted; `True`/`False` still coerce to `1`/`0`.
- Out-of-range value assignment (`2.0`) is still rejected.

`pytest test/test_result_nan_default.py test/test_result_bool.py` → 57 passed. `ruff check`/`ruff format` clean.

## Summary by Sourcery

Allow boolean result parameters to use NaN as the standard "missing" sentinel while preserving existing bounds and behavior.

Bug Fixes:
- Permit ResultBool to accept NaN as a default and assigned value without triggering bounds validation errors, while still rejecting truly out-of-range values.

Documentation:
- Document the ResultBool NaN handling fix in the unreleased changelog section.

Tests:
- Extend result NaN default tests to cover ResultBool defaults, subclass overrides, runtime NaN assignments, boolean coercion, and out-of-bounds rejection.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+- `ResultBool` rejected NaN as a default or value even though NaN is the documented "missing"/unrecorded sentinel for result variables (see `ResultFloat.__init__`). `ResultBool` locks its bounds to `[0, 1]`, and param validates a Parameter's default against its bounds whenever a subclass *overrides* it (and validates every value assignment). So `ResultBool(default=float("nan"))` raised `must be at most 1, not nan` the moment a subclass overrode the inherited Parameter, and assigning `float("nan")` at runtime to mark a sample missing raised the same — making the NaN "missing" sentinel that already works for `ResultFloat` unusable for `ResultBool`. `ResultBool._validate_bounds` now treats NaN as in-bounds, so result bools can use the same missing sentinel as `ResultFloat` while genuinely out-of-range values (e.g. `2.0`) are still rejected. Added coverage in `test/test_result_nan_default.py`.
+
 ## [1.104.1] - 2026-06-09
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


blooop:
Addressed Sourcery feedback in d3cd0fd:

- **Generalised the NaN check** in \`ResultBool._validate_bounds\` to use \`math.isnan\` (wrapped in try/except for non-numeric values) instead of \`isinstance(val, float)\`, so numpy NaN scalars (e.g. \`np.float32("nan")\`) are recognised too. Added a test covering this.
- **Added a lower-bound rejection case** (\`-1.0\`) to \`test_out_of_bounds_value_still_rejected\`.

Also fixed a related bug this PR surfaces: the \`ResultBool\` binomial SE (REDUCE path) divided \`p*(1-p)\` by the full repeat-dim size while computing \`p\` with \`skipna=True\`; with NaN now a valid "missing" repeat, the SE was understated. It now divides by the per-cell count of valid (non-NaN) repeats, with coverage in \`test/test_result_bool.py\`. Bumped to 1.104.2.

PR #960, opened 2026-06-12
Implement plan 05: test coverage gaps

blooop:
## Summary

Implements `plans/05-test-coverage.md` (from #959): direct unit tests for the previously indirectly-tested `results/` and core modules, de-flaked file server tests, and resolved long-skipped tests. One module per commit, as the plan prescribes.

**Coverage: 90% → 91%** (missed lines 1050 → 1001 of 10776). Tests: 1474 passed / 5 skipped → **1651 passed / 2 skipped** (+177 tests across 16 new test files).

### Task 1 — unit tests for untested result modules

| Module | Coverage before → after |
|---|---|
| `results/bench_result.py` | 75% → 83% |
| `results/histogram_result.py` | 83% → 83% (behavioral assertions added) |
| `results/optimize_result.py` | 97% → 100% |
| `results/dataset_result.py` | 64% → 86% |
| `holoview_results/scatter_result.py` | 76% → 100% |
| `holoview_results/bar_result.py` | 96% → 96% |
| `holoview_results/curve_result.py` | 100% (direct tests added) |
| `holoview_results/band_result.py` | 71% → 95% |
| `holoview_results/table_result.py` | 100% (direct tests added) |
| `distribution_result/` box/violin/jitter | 100% (direct tests added) |
| `results/volume_result.py` | 100% (direct tests added) |
| `composable_container_dataframe.py` | 100% (data-intactness tests added) |

Several of these were already at high *line* coverage purely through generated-example integration tests; the new tests assert behavior directly — element types, kdims/vdims, title/ylabel/units propagation, values matching the worker function, filter rejection paths, and NaN robustness (NaN is the missing-value default since v1.105).

### Task 2 — core modules

- `test_bench_cfg.py` (42 tests): defaults, flag round-trips, deprecated `level=` mapping, persistent hash behavior, describe/sentence helpers, optuna target partition. `bench_cfg.py` 97% → 99%.
- `test_worker_job.py` (16 tests): input assembly, canonical form, hash stability/sensitivity.
- `video_writer.py`: **no new file** — the plan's target (write path with synthetic frames, file exists and non-empty) is already covered by `test_video_writer_extended.py`.

### Task 3 — de-flake `test_file_server.py`

Replaced each `time.sleep(0.3)` with a `wait_for_port()` TCP poll (5s timeout, 0.1s step, `TimeoutError` on failure). Verified stable across 10 consecutive runs (all passed).

### Task 4 — long-skipped tests (with deviations; the plan's named tests don't exist)

- `test_combinations.py`: **not** an empty placeholder — a 184-line hypothesis class fully `@pytest.mark.skip` since 2023. Deleted per the plan's intent.
- The skipped test in `test_sweep_vars.py` is `test_int_sweep_samples_all` (not `test_missing_default_value`): an exact line-for-line duplicate of the passing test above it. Deleted.
- `test_bencher.py::test_unique_file_names` (not `test_plot_all_permutations`): bare skip replaced with a self-describing reason pointing at the plan.
- `test_usability.py`: deprecated `bn.ResultVar` → `bn.ResultFloat` (verified identical behavior).

### Bugs found while writing tests (not fixed here — test-only PR)

1. `ScatterResult.to_scatter` filters `result_types=(ResultVar,)`, but `ResultVar` is the deprecated shim — modern `ResultFloat` sweeps silently return `None` (the shipped scatter example appends `None` to its report). Pinned by `test_to_scatter_result_float_returns_none`; the filter likely wants `ResultFloat`.
2. `DistributionResult._plot_distribution` forwards `title` inside `**kwargs` *and* as an explicit kwarg into `.opts()` — any caller passing `title=` would hit "multiple values for keyword argument".
3. Histogram output never shows result-var units (only the var name); tests assert the implemented behavior.

### Notes

- `fail_under` (plan 01) is not in place yet, so there is nothing to raise; when plan 01 lands it can start at 91%.
- Full `pixi run ci` passes locally.

## Summary by Sourcery

Increase unit test coverage for core result and configuration modules and stabilize flaky file server tests.

Bug Fixes:
- Deflake file server integration tests by replacing fixed sleeps with port polling and clarify a previously bare skip reason.
- Remove redundant and long-skipped sweep/combinations tests and update usability tests to use the non-deprecated ResultFloat API.

Enhancements:
- Add focused unit tests for BenchRunCfg/BenchCfg behavior including defaults, hashing, and descriptive helpers.
- Introduce direct behavioral tests for BenchResult and various Holoviews-based result types (curve, bar, band, histogram, scatter, box/violin/jitter, volume, table, dataset).
- Add tests for WorkerJob input assembly and hashing and for DataFrame-based composable container datasets, including NaN robustness and data integrity.


blooop:
"Histogram output never shows result-var units (only the var name); tests assert the implemented behaviour"   this is not desired.  All graphs should show units. check this across the codebase (make a follow up pr if appropriate)

blooop:
Good catch — fixed the histogram in this PR (e9a91fc4): the x-axis now shows `{name} [{units}]` like band/bar/heatmap, and the test asserts `value [m]` instead of pinning the old behavior.

Audited the rest of the codebase per your request. Units are already shown by `band`, `bar`, `heatmap`, `surface`, and the distribution plots (box/violin/jitter inherit from `DistributionResult`). The remaining gaps are `curve_result.py`, `line_result.py`, and `scatter_result.py`, which don't set an axis label with units at all. I'll address those in a follow-up PR since they're production-only changes outside this test-coverage PR's scope.

Also addressed the Sourcery feedback: consolidated the duplicated `unwrap_hv` / `_inner_element` / `run_cfg_with` helpers into `test/helpers.py`, and added a BandResult negative test (the suggested `repeats=1` case isn't rejected by band's filter, so the test pins the real rejection path — a non-scalar vector result).

blooop:
Follow-up PR for the remaining axis-unit gaps (curve/line/scatter) is up: #963

PR #963, opened 2026-06-12
fix: show units on curve, line and scatter axis labels

blooop:
## Summary

Follow-up to #960, as promised in [this comment](https://github.com/blooop/bencher/pull/960#issuecomment-4692463977): the audit found that `curve_result.py`, `line_result.py` and `scatter_result.py` were the remaining plots that didn't show units on their axes (they set no axis label at all).

- Adds a shared `label_with_units()` helper in `bencher.utils` producing `name [units]`, following the existing `'ul'` = unitless convention from `sweep_base.describe_variable`
- Applies it to:
  - `HoloviewResult._build_curve_overlay` — covers `to_curve` and the aggregated line/over-time path (x = float input var, y = result var)
  - `LineResult.to_line_ds` (float-x path and 0D over_time path) and `_to_line_tap_ds`
  - `ScatterResult._to_scatter_ds`
- Labels are applied with `kwargs.setdefault`, so explicit user `xlabel`/`ylabel` kwargs still take precedence

## Test plan

- New `test/test_axis_units.py`: asserts `distance [m]` / `throughput [ops/s]` on curve and line axes, `score [m]` on scatter, and unit tests for `label_with_units` (units / no units / `'ul'`)
- Full suite: 1485 passed, 5 skipped; ruff + pylint clean

## Summary by Sourcery

Add shared axis labeling helper and ensure curve, line, and scatter plots display variable units on their axes.

New Features:
- Introduce a reusable label_with_units helper to generate axis labels from parameter names and units.

Enhancements:
- Apply unit-aware axis labels to curve, line, and scatter Holoviews plots while preserving user-provided labels.

Tests:
- Add axis unit tests for curve, line, and scatter plots, along with unit tests for the label_with_units helper.

### DIFF
--- a/bencher/results/holoview_results/holoview_result.py
+++ b/bencher/results/holoview_results/holoview_result.py
@@ -13,6 +13,7 @@ from bencher.utils import (
     get_nearest_coords1D,
     hmap_canonical_input,
     get_nearest_coords,
+    label_with_units,
     listify,
 )
 from bencher.results.pane_result import PaneResult
@@ -254,6 +255,12 @@ class HoloviewResult(PaneResult):
             return None
         kdims = [d for d in ds_dims if d in float_names] or ds_dims[:1]
         groupby = [d for d in ds_dims if d not in kdims]
+
+        # Show units on both axes: x from the float input var, y from the result var
+        kwargs.setdefault("ylabel", label_with_units(result_var))
+        x_var = next((fv for fv in self.plt_cnt_cfg.float_vars if fv.name == kdims[0]), None)
+        if x_var is not None:
+            kwargs.setdefault("xlabel", label_with_units(x_var))
 
         vdims = [var, std_var] if has_spread else [var]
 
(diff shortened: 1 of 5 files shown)


blooop:
Addressed the Sourcery review in a08c4720:

- **Tap-based line path coverage** — added `test_line_tap_axis_labels_show_units` exercising `to_line(use_tap=True)` (`_to_line_tap_ds`), asserting the same `distance [m]` / `throughput [ops/s]` labels via a shared `_assert_line_axis_labels` helper.
- **`label_with_units` edge cases** — added `test_no_units_attribute` (object with only `name`, exercising the `getattr(var, "units", "")` fallback) and `test_units_none` (`units=None`), both yielding the bare name.

On the two non-blocking suggestions, no change made:
- **`float_vars[0]` IndexError** — `to_line_ds` already guards with an early `return None` when `not self.plt_cnt_cfg.float_vars` (line 139), and in `_to_line_tap_ds` the `x = float_vars[0].name` access is pre-existing; this PR only adds the label line, so it introduces no new index risk. Out of scope to refactor here.
- **`.name` fallback in `label_with_units`** — all callers pass param-like objects that always have `.name`; keeping the helper strict avoids masking a genuine misuse.

PR #973, opened 2026-07-01
A1 Phase 3: named-only plugins — every result type addressable by name in to_auto

blooop:
## Summary

A1 Phase 3 (revised per the addendum in `plans/architecture/A1`): **every result type is now addressable by name in `to_auto`, without changing what a default report renders.**

The mechanism is a *named-only* plugin concept: plugins registered with `auto=False` live in the registry (addressable, listable, overridable) but are skipped during automatic selection; naming them via `plot_list`/`include`/`only` selects them like any other plugin. The attribute is optional on the `PlotPlugin` protocol (read with a `getattr` default), so every existing plugin — including third-party ones — is unaffected.

```python
res.to_auto(plot_list=["violin"])          # now works by name
res.to_auto(plot_list=["line", "table"])   # mix default + named-only chart types
```

## What's registered

| Chart type | Backend | Callback |
|---|---|---|
| violin, scatter_jitter, scatter, band, table | holoviews | `*.to_plot` |
| tabulator, dataset, video_summary | panel | `to_plot` / `to_video_summary` |
| surface | plotly | `SurfaceResult.to_plot` — Plotly only where a plot already required it (like the volume built-in); no Plotly port |
| rerun | **rerun** | `RerunResult.to_rerun` — Rerun as its own first-class backend; the SDK is imported lazily inside the renderer, so registration needs no install guard |

## Also fixed

`LegacyResultPlugin.render` now filters the ride-along kwargs (`override` + plot-size) to the callback's signature when the renderer has no `**kwargs` — previously a fixed-signature renderer like `RerunResult.to_rerun` would TypeError and be dropped.

## Verification

- [x] Default output unchanged: the Phase 2 parity tests are untouched and green; new test asserts named-only plugins never appear in automatic selection
- [x] `pixi run format` / lint / pylint 10.00/10
- [x] Full suite: 1737 passed, 2 skipped (includes 12 new tests: registry-level `auto` semantics, backend assignments, violin/table by name, name↔callable equivalence, composing named-only with the auto set, kwargs filtering, plugins without the `auto` attribute staying automatic)
- [x] Split-render guard tests (`test_render.py`, `test_split_render_examples.py`): 50 passed

## Next

A2 (plot-selection redesign): lift the dynamic `to_plot`-internal filters into declarative registry signatures, ranking, and serializable plot specs — that's where named-only chart types gain shape-aware auto-selection if desired.

## Summary by Sourcery

Introduce named-only plot plugins so every result type can be selected by name in to_auto without changing default report output, and refine plugin behavior and docs accordingly.

New Features:
- Add named-only plugin support via an optional auto=False attribute on plot plugins, making them selectable only when explicitly named in plot_list/include/only.
- Register non-default built-in chart types (violin, scatter variants, band, surface, table, tabulator, dataset, video_summary, rerun) as named-only plugins addressable by name in to_auto.

Bug Fixes:
- Ensure LegacyResultPlugin.render filters render kwargs to the callback signature for fixed-signature renderers to avoid TypeError.

Enhancements:
- Extend the plot_plugin decorator and internal plugin types to carry an auto flag while keeping older plugins automatically selected by default.
- Update plugin selection logic to skip named-only plugins in automatic selection while still honoring them when explicitly requested.
- Document the named-only plugin behavior and the auto flag in the plot plugin design and A1 architecture plan, including backend assignments for new built-ins.

Build:
- Bump project version to 1.112.0 and record the new named-only plugin capability and kwargs filtering behavior in the changelog.

Documentation:
- Document the auto flag and named-only plugin semantics in the plot plugin design doc and A1 architecture plan.

Tests:
- Add tests covering named-only built-in plugins, backend assignments, auto-selection exclusion, selection by name equivalence, composition with the auto set, kwargs filtering for fixed-signature callbacks, and behavior of plugins without the auto attribute.


blooop:
Both Sourcery points addressed in 2d6e4820:
- `inspect.signature` is now behind an `lru_cache`'d module helper (`_declared_kwargs`) — one inspection per callback ever, render-loop cost is a dict/set intersection only for fixed-signature callbacks.
- Added `test_kwargs_callback_gets_full_render_kwargs`: a `**kwargs` renderer receives the render_kwargs dict unfiltered, including undeclared names (guards the VAR_KEYWORD branch against accidental filtering).

blooop:
## Review (8-angle fan-out + adversarial verification)

**Verdict: approach is sound; one confirmed user-facing bug found and fixed on the branch.**

### Confirmed bug (fixed in 9b29e4c2)
- `to_auto(plot_list=["tabulator"])` **always failed**: `TabulatorResult` was never in `BenchResult`'s base list, so the unbound callback hit `AttributeError: 'BenchResult' object has no attribute 'to_tabulator'` — swallowed into the "No Plotters" fallback. `table` only worked by accident (its `to_plot` uses `HoloviewResult` methods). Fixed by adding `TableResult`/`TabulatorResult` to the MRO (late, so they shadow nothing), verified end-to-end. Added a test that renders `tabulator` by name, plus a structural test asserting every named-only callback's defining class is in `BenchResult`'s MRO (catches the latent trap for future named-only specs).

### Robustness (fixed in same commit)
- `_declared_kwargs`: dropped `lru_cache(maxsize=None)` — it pinned every callback (incl. per-run closures/bound methods and their captured datasets) for process lifetime and crashed on unhashable callables; it now also returns `None` (no filtering) for non-introspectable callables (C builtins) instead of raising. Signature inspection is ~µs once per plugin per report; the cache wasn't paying for itself.

### Merge conflict resolution (1373479d)
- main took the 1.112/1.113 series, so this PR is re-versioned to **1.114.0** and the changelog entry moved above 1.113.1.

### Noted, not blocking (follow-up material)
1. `include` runs the match filter but `only` bypasses it — a named-only plugin with an honest (non-`match_all`) filter is reachable via `only` but silently dropped via `include`, contradicting the docstrings. Collides with A2's plan to lift `to_plot`-internal filters into the registry — resolve there.
2. `getattr(p, "auto", True)` sniffs a generic attribute name off duck-typed third-party plugins; a plugin with an unrelated `auto=False` attribute is silently demoted to named-only.
3. The rerun import guard in `bench_result.py` (try/except + stub) is now dead protection: `_named_only_specs` imports `RerunResult` unguarded at registration time.
4. `register_builtin_plugins` has two near-identical registration loops sharing an implicit priority counter — could fold into one loop with an `auto` flag on the spec tuples.
5. Kwargs filtering means a typo'd kwarg for a fixed-signature renderer is now silently dropped instead of raising — inherent trade-off of the fix, worth remembering in A2's declarative-signature work.

Local verification: plugin suites 62/62; full suite 1834 passed (2 pre-existing `test_result_collector` dtype failures reproduce on origin/main in this env, unrelated to the PR).

PR #976, opened 2026-07-05
plans: result cache & history invalidation correctness (research + fix prompt)

blooop:
## What

Adds `plans/09-result-cache-invalidation.md` — a self-contained task prompt for a follow-up agent/developer to **research and fix** two correctness defects in how `BenchCfg.hash_persistent` invalidates the result cache (Layer B) and the `over_time` history cache (Layer C). No code change here; this is a plan doc in the existing `plans/` convention (+ one README index row).

## The two defects it documents

**D1 — `result_vars` *order* changes the key (should be a no-op).**
`hash_persistent` folds each var into a running accumulator *in list order* (`bench_cfg.py:796-798`). Permuting `result_vars` — which changes nothing about *what* is measured — changes the hash whenever the moved vars differ in class/units/direction, so both Layer B and Layer C miss and the benchmark's entire history + regression baselines are silently discarded. The mental model (and the determinism tests) treat the metric set as a *set*; nothing documents that reorder is a full invalidation.

```python
a = bn.ResultBool(units="ratio", direction=OptDir.maximize)
b = bn.ResultFloat(units="s", direction=OptDir.minimize)
assert key([a, b]) == key([b, a])   # fails today
```

**D2 — the per-variable hash ignores the variable *name*.**
`_hash_slots` (`variables/results.py:50`) hashes class + `__slots__` (units, direction, bounds) but stops at the param framework, so `name` is never hashed. Renaming a result var therefore does **not** change the key — but the history dataset's data variables are named by `rv.name`, so `xr.concat([ds_old, dataset], "over_time")` outer-joins old-name vs new-name into a phantom dead column + an apparently-reset series. A silent partial corruption, worse than a clean reset.

```python
x = bn.ResultFloat(units="s", direction=OptDir.minimize); x.name = "duration"
y = bn.ResultFloat(units="s", direction=OptDir.minimize); y.name = "latency"
assert x.hash_persistent() != y.hash_persistent()   # fails today
```

Plus **D3** — both invalidation paths are silent (info/debug log only), so an accidental reset is indistinguishable from an intended one.

## Scope / coordination

The doc includes reproductions, file refs, research questions (can `result_vars` be made order-independent while `input_vars` stays ordered? should name be part of identity?), a proposed order-independent + name-inclusive key with load-time reconciliation, and acceptance criteria (new invariant tests, determinism contract preserved, `CACHE_VERSION` bump + CHANGELOG).

It explicitly coordinates with `plans/architecture/A4-caching-architecture.md`, whose §5 declines to change `hash_persistent` semantics — A4's guidance targets the intentional `title` exclusion; this plan scopes the ordering/naming behavior as a separate, genuine defect (implement inside A4's `caching/keys.py` if that lands first).

## Summary by Sourcery

Add a new planning document describing how to fix correctness issues in benchmark result and history cache invalidation, and index it in the plans README.

Documentation:
- Introduce Plan 09 documenting defects and proposed fixes for result cache and history invalidation semantics, including research questions and acceptance criteria.
- Update the plans index to reference the new cache and history invalidation correctness plan.

### DIFF
--- a/plans/09-result-cache-invalidation.md
+++ b/plans/09-result-cache-invalidation.md
@@ -1,23 +1,36 @@
 # Plan 09 — Result Cache & History Invalidation Correctness
 
-**Goal:** Fix two correctness defects in how the benchmark-config hash
+**Goal:** Fix three correctness defects in how the benchmark-config hash
 (`BenchCfg.hash_persistent`) invalidates the result cache (Layer B) and the
 `over_time` history cache (Layer C):
 
-1. It is **sensitive to the order** of `result_vars`, so reordering the result
-   columns — a change that alters nothing about *what* is measured — silently
-   discards a benchmark's entire accumulated history and regression baselines.
-2. The per-variable hash **excludes the variable name**, so *renaming* a result
-   variable does not change the key, yet the stored history is keyed internally by
-   name. The result is a silent, partial history split rather than a clean reset.
+1. It is **sensitive to list order** — for `result_vars` *and* `const_vars` — so
+   reordering the result columns, or reordering a dict of constants, silently
+   discards a benchmark's entire accumulated history and regression baselines,
+   though nothing about *what* is measured changed.
+2. The per-variable hash **excludes the variable name** — for result vars *and*
+   input/sweep vars — so *renaming* a variable does not change the key, yet the
+   stored history is keyed internally by name: silent, partial history
+   corruption (a different flavor per variable kind — see D2).
+3. Invalidation is **silent**: when the key moves, history vanishes with only an
+   info-level log. D3 proposes a last-seen-key index and an `on_history_reset`
+   policy knob to make resets observable.
+
+These semantics demonstrably violate user expectations: large downstream
+benchmark suites document the key in their own contributor docs as "a hash of
+the *set* of input/result/const vars" (it is not — it is an ordered fold), and
+invent conventions such as routing every result-var list through one shared
+constant purely to protect against resets they cannot see. When users build
+folklore defenses around a cache key, the key should change, not the users.
 
 **⚠️ Read first:** This plan deliberately revisits `hash_persistent()` *semantics*,
 which `plans/architecture/A4-caching-architecture.md` §5 says NOT to change. A4's
 "don't touch it" refers to the intentional exclusion of display-only fields (e.g.
-`title`) — that guidance stands. The ordering/naming behavior described here is a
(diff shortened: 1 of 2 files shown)


blooop:
Updated the plan doc (204 → 283 lines) after re-verifying every claim against current source. Two factual corrections and three extensions:

**Corrections**
- The `over_time` history cache (Layer C) is keyed by `hash_persistent(True)` — the *same* hash as the result cache — not `include_repeats=False`. `hash_persistent(False)` keys only the sample cache. (`bencher.py:678,682,727`; repeats are intentionally in the history key so historical arrays share a shape.) The stale wording also lives in A4 §2 W5 — flagged in the doc.
- Per-variable identity is effectively `(class, units)`, not `(class, units, direction)`: `direction` (with `share_axis`/`max_time_events`) is in `_hash_exclude` (`variables/results.py:107`). The old D2 repro passed `direction=` kwargs that don't affect the hash — removed.

**Extensions**
- **D1 now covers `const_vars`** — folded in list order too (`bench_cfg.py:800-801`), and const order comes from `list(dict.items())` / class-declaration order, affecting nothing but the title string. (The const *value* in the key is intentional.)
- **D2 now covers input/sweep vars** — `SweepBase._sweep_identity` never includes `self.name`. Renaming an input var is *worse* than a result var: instead of a NaN dead-column split, `xr.concat` broadcasts into a phantom dimension — fabricated data points, zero NaNs to betray it (executable verification in the doc, checked on xarray 2025.6.1).
- **D3 gained a concrete remediation** — a per-`(bench_name, tag)` last-seen-key index + identity summary, a WARNING classifying the change and counting orphaned `over_time` events, and `on_history_reset = "warn"|"error"|"ignore"` (default `"warn"`) on `BenchRunCfg`.

Order-independence is explicitly scoped to `result_vars` + `const_vars`; `input_vars` ordering is left as research question 3 (dim-layout/positional-code evidence cited).

blooop:
Addressed the Sourcery feedback in 179db00:
- **§4 combiner:** now spells out sorted-tuple vs commutative/multiset hashing and picks sorted-tuple, with the rationale (XOR cancels duplicates and same-type vars already share an identity, so a commutative combiner could silently zero them out).
- **D3 surfacing:** now concrete — a `logging.WARNING` (up from today's INFO/DEBUG), a note in the rendered report's over-time/history section (the A4 §3.5 surfacing), and the `on_history_reset` knob (`warn` emits both; `error` raises `HistoryResetError`; `ignore` suppresses).

PR #977, opened 2026-07-06
plans: regression policy on result variables & structured verdict export (research + fix prompt)

blooop:
## What

Adds `plans/10-regression-policy-and-verdict-export.md` — a self-contained research+fix plan (no code change) in the `plans/` convention, plus one README index row.

## Problem

Regression policy today lives far from the metrics it governs: per-variable thresholds are passed run-side via `regression_overrides`, and the verdicts leave the library only as rendered report content or ad-hoc consumer parsing. CI systems that gate merges on "did metric X regress" maintain out-of-band `benchmark/metric → threshold` registries and re-serialize `bench.results[*].regression_report` into hand-rolled text formats. There is no concept of *severity* (hard-gate vs notify-only), so users encode it in parallel registries and parallel output files.

## Direction the plan develops

- **Declare regression policy on the result variable** via a `RegressionSpec` that normalizes to the same `{method: threshold}` shape as the existing `regression_overrides` (builds on #974, not around it). Class-level declaration travels with the metric across every benchmark that reuses the var; run-level overrides remain the per-run layer, with defined precedence.
- **Severity as a first-class field** (`gate`/`notify`/`info`), default `gate` so existing `regression_fail` behavior is byte-identical.
- **Structured verdict export** extending the existing `result_to_dict` / `RegressionReport.to_dict` (adds `severity`, `units`, `has_gate_regressions`, `severity_counts`) — no parallel artifact — plus a `has_gate_regressions` accessor so CI needn't parse for the common case.
- **Publisher awareness** via a `BenchReport.verdict_summary()` accessor rather than widening the `publish()` signature.

## Notes

- Load-bearing subtlety flagged in the plan: the new `regression` slot on result vars **must** be added to `_hash_exclude`, or tightening a threshold would wipe the very over_time history the detector compares against. Acceptance criteria include a hash-stability test.
- Coordinates with plan 08 (the detection/rendering split in `regression.py`) — this plan is policy declaration + export only — and with plan 09 / A4 (avoids hash-key semantics, only stays out of the hash).
- README index row inserted after the 08 row; if PR #976 (plan 09) merges first, the row placement is a trivial one-line conflict.

## Summary by Sourcery

Add a new plan describing regression policy on result variables and structured verdict export, and index it in the plans README.

Documentation:
- Add plan 10 document outlining regression policy on result variables, severity handling, and structured verdict export, including background, design, phases, and acceptance criteria.
- Update plans README to reference the new regression policy and verdict export plan with its scope and ordering guidance.

### DIFF
--- a/plans/10-regression-policy-and-verdict-export.md
+++ b/plans/10-regression-policy-and-verdict-export.md
@@ -26,6 +26,11 @@ landed, the same symbols simply live in two files; the changes below are unaffec
   this is a hard requirement, enforced by the slot-coverage contract).
 - If a step fails in a way this plan does not cover, stop and report rather than
   improvising.
+- **Line numbers are as-of the current `main` (CHANGELOG ≤ 1.113.0) and will
+  rot** — the durable reference is always the named symbol (`detect_regressions`,
+  `_normalize_overrides`, `RegressionResult.to_dict`, `_hash_exclude`, …). If a
+  cited line has moved, grep the symbol; the responsibility, not the address, is
+  what the plan depends on.
 
 ---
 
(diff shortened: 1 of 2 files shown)


blooop:
**Coordination with open PR #941 (`bencher.ci` module).** #941 extracts `write_performance_summary()` / `render_regression_plots()` / `warn_on_regressions()` — the CI-serialization side of this same problem. This plan is complementary but overlapping and the two should be reconciled before implementing:

- #941 serializes a `RegressionReport` to a **pipe-delimited** text file with per-metric filtering/thresholds passed in by the caller.
- This plan proposes declaring policy (+ **severity**) **on the result variable**, and a **structured JSON** verdict export extending `result_to_dict` with a `has_gate_regressions` accessor.

The severity concept and the structured export would let `bencher.ci` drop the external threshold-registry argument and the hand-rolled pipe schema (or keep the pipe writer as one renderer over the structured verdicts). Suggest deciding whether #941 lands first as the plumbing and this plan layers policy/severity on top, or this plan's export becomes the source `bencher.ci` serializes from.

blooop:
Addressed the Sourcery feedback in 19747a4:
- **Precedence/severity in one place:** added §4.0 'Semantics at a glance' — the authoritative precedence order (run override ▸ class spec ▸ benchmark-wide, including `{}` opt-out, replacement-not-merge), plus a table of how `gate`/`notify`/`info` interact with `has_regressions`/`has_gate_regressions`/`regression_fail` for mixed-severity variables.
- **Line-number rot:** added a Rules note that line numbers are as-of current `main` and will rot; the named symbol is the durable reference. Kept the file:line refs (consistent with plans 01–08 and useful to an implementer now), but every ref already names its symbol.

PR #980, opened 2026-07-06
plans: benchmark declaration bundle & run defaults (research + fix prompt)

blooop:
## What

Adds `plans/13-benchmark-declaration-and-run-defaults.md` — a self-contained research+fix plan (no code change), plus one README index row.

## Problem

A benchmark function carries no defaults for repeats, subsampling, tag, over_time, `max_time_events`, or its required sampling context, so every entry point that runs it (a `__main__` block, a test wrapper, a CLI, a batch runner) re-specifies them and they drift apart in large suites. Unattended/CI loops reinvent env overrides around every `__main__`. Wrappers that want "tag defaults to the function name" resort to `inspect.currentframe().f_back` hacks. Scorecard grouping and result-var display aliases are maintained in external registries.

## Direction the plan develops

- **`@bn.benchmark(...)` declaration bundle** — a frozen `BenchDeclaration` attached to `fn.__bench_declaration__`, returning the function unchanged so existing signature-sniffing keeps working; `bn.get_declaration()` lets harnesses read intent without running. `sampling_context` declared as a zero-arg factory (context managers are single-use).
- **Single precedence spine** `env > call-site > declared > library default`, reusing the repo's existing `UNSET` sentinel to disambiguate call-site from default.
- **First-class env overrides in `bn.run`** (`BENCHER_REPEATS`, `BENCHER_SHOW=none` instead of a competing HEADLESS knob), strict parsing; `BenchRunner` stays env-free for deterministic composition.
- **Tag derivation** from the callable `__name__` — gated on the decorator, because `run_tag` flows into the persistent hash and auto-tagging undecorated benchmarks would silently re-key every existing cache.
- **`category=`/`order=` report metadata** for native scorecard grouping and a render-time **`display_name=`** on result vars (in `_hash_exclude`, so identity/history are unaffected).

## Notes

- Explicit coordination note with plan 09: whichever lands second must keep `display_name` out of any name-in-hash change.
- README index row inserted after the 08 row; trivial one-line conflict if earlier plans merge first.

## Summary by Sourcery

Add a new research plan describing benchmark declaration bundles and run-defaults, and index it in the plans README.

Documentation:
- Add Plan 13 document detailing benchmark declaration bundle, run defaults, env overrides, tagging, metadata, and phased implementation steps.
- Update plans README to reference Plan 13 and note coordination with Plan 09.

### DIFF
--- a/plans/13-benchmark-declaration-and-run-defaults.md
+++ b/plans/13-benchmark-declaration-and-run-defaults.md
@@ -138,8 +138,11 @@ Resolved once, at the top of `bn.run` (the canonical unattended entry point).
 
 Parsing is **strict**: a set-but-unparsable value raises `ValueError` naming
 the variable and its accepted values — a silently ignored override in CI
-produces wrong data, worse than a crash. Empty string means unset (supports
-`BENCHER_REPEATS= cmd`); bools accept `1/0/true/false/yes/no`
+produces wrong data, worse than a crash. An empty value counts as **unset**
+(falls back to the declared/library default), so a per-invocation assignment
+prefix such as `BENCHER_REPEATS= python -m my_benchmarks` — a POSIX var-assignment
+prefix, no space after the `=`, setting the variable to empty for that one
+command — is a no-op rather than an error. Bools accept `1/0/true/false/yes/no`
 (case-insensitive); `BENCHER_SHOW` goes through `normalize_show`
 (`bencher/bench_cfg.py:44`) plus the bool spellings. No lenient mode.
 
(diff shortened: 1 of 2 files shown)


blooop:
Addressed the Sourcery inline comment in 0160131. The original `BENCHER_REPEATS= cmd` is actually valid POSIX (an assignment prefix with no space after `=`), but it reads ambiguously. Reworded it to an explicit example — `BENCHER_REPEATS= python -m my_benchmarks` — and stated inline that it's a var-assignment prefix (no space after the `=`) setting the variable empty for that one command, so it's a no-op fallback rather than an error.

PR #993, opened 2026-07-29
feat: add xy_histogram and xy_hexbin chart types

blooop:
Stacked on #992. Two more intra-sample charts on the `TabularSpec` base, both answering "where is the mass in this sample's rows?" — in one and two dimensions.

## `xy_histogram`

Bins measured columns of a `ResultDataSet`. The existing `histogram` bins a `ResultFloat` *across* the sweep, so what it shows is the spread of the repeats, not the distribution a single sample measured (every request timed, rather than the mean).

- `column=` takes a list to overlay several, **binned over a shared range** so they're actually comparable — binned to their own spans they would look alike regardless of where they sit — and drawn translucent so the lower one stays readable.
- Defaults to every numeric column.
- `density=True` normalises instead of counting, and relabels the y axis accordingly.
- An empty or all-NaN column produces **empty bins rather than raising**: numpy can't pick a range for an empty array, and NaN is how bencher marks a sample missing, so NaN rows are dropped rather than poisoning the range.

## `xy_hexbin`

The density counterpart to `xy_scatter`, same axes and options. At tens of thousands of points the markers saturate, and the shape of the distribution is precisely what a scatter then loses. `gridsize=` sets resolution, `min_count=1` drops empty tiles, and the colourbar is on by default since density without a scale shows shape but no magnitude.

## Also

`hv.HexTiles` added to `HoloviewResult.DEFAULT_SIZED_ELEMENTS` — it was absent, so a hexbin would have taken the holoviews default size instead of bencher's 600x600. Same gap that was fixed for Histogram/Area/ErrorBars previously, and the existing coverage test picks it up automatically.

## Rebase note (2026-07-29)

Rebased onto #992's rebased tip. `TabularSpecResult` is gone from #991 after review, so `XYHistogramResult` and `XYHexbinResult` subclass `HoloviewResult` and compose `render_data_samples` — the same per-sample path `PaneResult.to_panes` and the default report take — instead of inheriting a family-specific result base. Both `build()` implementations, `resolve_columns`, and the tests are unchanged.

## Tests

`test/test_xy_distribution_results.py`, 46 tests, both charts sharing fixtures: factories, the shared bin range, translucency, missing/empty data, both chart types, plugin registration, pickling, both declarative routes.

## Examples

`example_plot_xy_histogram` (4000 timed requests per sample, measured vs baseline) and `example_plot_xy_hexbin` (20000 points, where a scatter stops working).

`pixi run ci` clean; full suite 2077 passed.

## Summary by Sourcery

Introduce intra-sample histogram and hexbin chart types for tabular results and wire them into the plotting, plugin, and example infrastructure.

New Features:
- Add xy_histogram chart type to bin one or more measured columns of a ResultDataSet within a single sample.
- Add xy_hexbin chart type as a hex-binned density counterpart to xy_scatter for dense point clouds.

Enhancements:
- Share column resolution logic for single-axis tabular charts via resolve_columns in TabularSpec.
- Register hv.HexTiles for the shared default Holoviews figure size and expose new xy chart types through BenchResult, plugins, and the public API.
- Provide new example benchmark configurations and gallery examples demonstrating xy_histogram and xy_hexbin usage.

Documentation:
- Document xy_histogram and xy_hexbin usage, options, and selection guidance in the user guide.
- Update changelog with details of the new intra-sample histogram and hexbin chart types.

Tests:
- Add comprehensive test suite covering xy_histogram and xy_hexbin factories, plugin registration, result rendering paths, pickling, and handling of missing/empty data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Changed
+- **Intra-sample chart gallery examples now declare their container** rather than appending
+  a report-level plot. `res.to(XYCurveResult, ...)` and friends *add* a plot below whatever
+  `plot_sweep` already rendered, which for a `ResultDataSet` with no declared container is
+  the raw table — so the example showed the rows and the chart. Declaring
+  `container=bn.xy_curve(...)` puts the chart in the result's own position instead, which is
+  what `example_plot_xy_scatter` already did. Affects `example_plot_xy_curve`,
+  `example_plot_xy_histogram` and `example_plot_xy_hexbin`; the chart-type route is
+  unchanged and still covered by tests.
 - **`DataSetResult` now renders `ResultDataSet` results only.** It previously claimed every
   pane-type result, which made `bench.add(bn.DataSetResult)` / `plot_list=["dataset"]` a
   second name for the `panes` view on a sweep with no stored payload. Now that
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   rather than indexing a list with an array several frames later.
 
 ### Added
+- **`xy_histogram` chart type** — bins one or more measured columns of a `ResultDataSet`,
+  showing the distribution a single sample measured. Distinct from the existing
+  `histogram`, which bins a `ResultFloat` across the sweep and so shows the spread of the
+  repeats. `column=` takes a list to overlay several distributions, binned over a *shared*
+  range so they are actually comparable, and drawn translucent so the one underneath stays
+  readable; it defaults to every numeric column. `density=True` normalises instead of
+  counting, and the y axis is labelled accordingly. An empty or all-NaN column produces
+  empty bins rather than raising — numpy cannot pick a range for an empty array, and NaN is
+  how bencher marks a sample missing, so NaN rows are dropped rather than poisoning the
+  range. Gallery example: `example_plot_xy_histogram`.
+- **`xy_hexbin` chart type** — the density counterpart to `xy_scatter`, taking the same
+  axes and binning them into hexagonal tiles. For a cloud of tens of thousands of points
+  the markers saturate and where the mass actually is becomes the thing a scatter cannot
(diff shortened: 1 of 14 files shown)


blooop:
this shows the raw table still. I just want the graph

blooop:
Pushed fadc69ba.

**"this shows the raw table still. I just want the graph"** — fixed. The examples declared no `container=` and appended `res.to(XYHistogramResult, ...)`, so the report got both: `plot_sweep` renders the `ResultDataSet` in the normal result position, which with no declared renderer is the raw table, and the chart landed below it. They now declare the container, which puts the chart *in* that position — the pattern `example_plot_xy_scatter` already used. Verified by walking the built report: `xy_histogram`, `xy_hexbin` (and `xy_curve`, below) each go from 1 table + 1 chart per sample to 0 tables, 1 chart.

I fixed `example_plot_xy_curve` the same way even though it merged in #992 — identical defect, same generator table, and the gallery would otherwise be inconsistent. Flagging it since it is outside this PR's original diff.

The `bench.add(bn.XYHistogramResult, ...)` route is unchanged and still covered by tests; `docs/how_to_use_bencher.md` now states plainly that a declared container *replaces* the table while a chart type *appends* below whatever `plot_sweep` rendered, which is the distinction that produced this.

**@sourcery-ai on label precedence** (both overall comments) — no change needed; `TabularSpec.element_opts` already resolves this the way you are asking for. `xlabel`/`ylabel` are passed to it as *defaults*:

```python
if (label := self.ylabel if self.ylabel is not None else ylabel) is not None:
    opts["ylabel"] = label
```

so `count`/`density` and the column names only apply when the spec left the field unset, and `opts.update(self.opts)` runs last so any raw holoviews keyword wins over the chart's own options. `gridsize`/`cmap`/`colorbar` are declared fields on the spec, so the value there *is* the user's. It was untested for these two charts, though, which is a fair thing to have flagged — five tests now pin it: explicit labels beating the density-derived default, that surviving an overlay, `opts` beating the overlay alpha, explicit hexbin labels, and `opts` applying after the chart options.

**Typo** — the clause parses (`the point at which [where the mass actually sits] becomes...`) but `is becomes` is a genuine stumble, so it is now `sits becomes`. Your suggested edit dropped `at which`, which breaks the sentence.

`pixi run ci` clean: 2116 passed, 3 skipped, pylint 10.00/10.

PR #994, opened 2026-07-29
feat: extend declared containers to every renderable result type

blooop:
**Branches off `main`, not the chart stack** (#991–#993) — it only touches `variables/results.py` and `bench_result_base.py`, so it has no dependency on that work and shouldn't queue behind three PRs.

## The asymmetry

`ResultDataSet` and `ResultReference` could declare how they render; the other renderable types could not. And `ResultReference` only honoured a container attached to a *stored sample* — one declared on the class was silently ignored, because `ds_to_container` consulted only `ref.container`, never `result_var.container`.

## What changes

`ResultString`, `ResultPath` and `ResultContainer` (so `ResultRerun` too, via inheritance) get the same declared-renderer slot, and **a declared container beats the type's built-in `to_container()`**:

- a `ResultPath` can render a file's *contents* — a CSV as a chart, a JSON as a tree — where before it was always a download widget;
- a `ResultString` can render as Markdown or highlighted code instead of plain text.

Precedence is uniform: renderer-supplied → the sample's → the class's → the type's default. Resolution is now one helper, `declared_container`, instead of being open-coded per type, and it uses `getattr` so a result pickled before a type gained the slot still renders.

## One breaking change

**A `ResultReference` container is now called with the object alone**, matching the `ResultDataSet` contract, so one renderer serves both. It was called as `container(obj, **kwargs)`, leaking in the render keywords every path adds (`override`, `agg_over_dims`, plot-size, `pane_layout`) — which means a single-argument callback raised `TypeError`. Breaking only for a callback that *relied* on receiving those keywords; nothing in bencher did.

The `container=` a renderer passes to `to_panes` is a **separate contract** — a panel pane constructor, still given `styles=` and the layout keywords, as `PaneResult.to_panes` and `video_container` require — and is deliberately untouched. Keeping these two distinct is the main thing worth reviewing here.

## Cache safety

Every added slot is in `_hash_exclude`, which `_hash_slots` drops before hashing, so declaring a renderer leaves cache keys and `over_time` history **byte-identical**. Verified against hashes recorded before the change, and asserted in `TestPersistentHashUnaffected`.

## Scope note

Deliberately **not** extended to `ResultVolume`, despite it looking like the same one-line addition: nothing dispatches on it. `VolumeResult` plots `ResultFloat` over three float inputs, and `ResultVolume` is absent from `PANEL_TYPES`, so it never reaches `ds_to_container` — a container slot there would be an option that never fires.

## Tests

`test/test_declared_container.py`, 14 tests: the helper (including a source lacking the slot), each newly-supported type with and without a declared container, that `to_container()` still wins when nothing is declared, per-sample overriding class-level, the single-argument call, and hash invariance.

`pixi run ci` clean; full suite 1935 passed; also verified under `BENCHER_FORCE_SPLIT_RENDER=1`.

## Summary by Sourcery

Extend declared container support across renderable result types and unify container resolution and calling conventions.

New Features:
- Allow ResultString, ResultPath and ResultContainer (and ResultRerun via inheritance) to declare custom render containers that override the type’s default rendering.
- Introduce a shared declared_container helper to resolve renderer-supplied, per-sample and class-level containers in a consistent precedence chain.

Bug Fixes:
- Ensure ResultReference honours containers declared on the class as well as per-sample, fixing previously ignored class-level declarations.
- Call ResultReference containers with the stored object alone instead of leaking render kwargs, aligning with ResultDataSet and avoiding TypeError for single-argument callbacks.

Enhancements:
- Keep declared container slots out of persistent hash computation so caches and over_time history remain unchanged when renderers are added.
- Clarify and document the distinction between renderer-supplied pane constructors and declared single-argument containers in code comments and user docs.

Documentation:
- Update usage documentation to list all result types that accept declared containers, describe their precedence, and provide examples for richer string and path rendering.

Tests:
- Add a dedicated test suite for declared container precedence, ResultString/ResultPath/ResultContainer/ResultReference behaviour, single-argument safety, and hash invariance.


CHANGELOG.md
```diff
@@ -8,6 +8,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Changed
+- **A `ResultReference` container is now called with the object alone.** It was called as
+  `container(obj, **kwargs)`, so the render kwargs every path adds (`override`,
+  `agg_over_dims`, the plot-size and `pane_layout` keywords) leaked into a callback that
```
blooop: Good catch that the phrasing was vague, though `plot_size` is not the right substitution — it is a `BenchCfg` field, not a kwarg the callback receives. `set_plot_size` reads `bench_cfg.plot_size` and injects `width`/`height`, so those are the keywords that actually leaked. Reworded to name them explicitly in f274a5e.

blooop:
Follow-up on the scope note: investigated `ResultVolume` and it is worse than "nothing dispatches on it" — it *crashes*. Declaring one raises `KeyError: No variable named ...` in `precompute_result_arrays`, because `setup_dataset` gives it no data variable but the collector indexes one regardless. So a container slot there would not merely never fire; the type cannot appear in `result_vars` at all.

Deleted in #1009 (independent, off `main`), with a guard test pinning that membership in `ALL_RESULT_TYPES` implies the collector can store the type. This PR is unchanged.

PR #999, opened 2026-07-29
docs: add plan 15 — stable benchmark series identity

blooop:
Adds `plans/15-benchmark-series-identity.md`. Plan doc only — no code changes.

Plan 09 shipped a last-seen index so that a moved history key is *reported*
rather than silently starting a new trend. That index is keyed by
`last_seen_key(bench_name, tag)` (`bencher/history.py:180`) — and both of those
are themselves inputs to the history key (`bencher/bench_cfg.py:831`, `:834`).
So the detector splits in two:

| What changed | Index entry | Detected? |
|---|---|---|
| `input_vars`, `const_vars`, `repeats` | stays put | **yes** — `full_reset` with a diff |
| `bench_name` or `tag` | **moves with the key** | **no** — miss on both, silent orphan |

A rename is the most likely way a key moves in practice, and it is exactly the
case that produces no event. Compounding it, `bench_name` defaults to the worker's
class name (`bencher/factories.py:39-42`) and is hashed, so renaming a class or
extracting a shared base silently resets history — the same hazard plan 13's D3
deliberately gates for `__name__`-derived tags.

The plan proposes `series_id` (defaulting to `f"{bench_name}:{tag}"`, excluded
from the hash) as the index key, and uses the `config_summary` already stored
there to classify a moved key: identical summary → **adopt** the record and report
`history_renamed`; differing summary → today's `full_reset`. So a pure rename
stops costing a benchmark its trend, and a real change stays reported.

Phased 1→5, each independently shippable, with phase 1 landing a golden-hash guard
before any behavior changes. Every claim is cited to file:line against `main` @
`7dad0cd4` (v1.116.0).

## Summary by Sourcery

Documentation:
- Add Plan 15 document describing series identity (`series_id`), its interaction with benchmark history, cache tags, and phased rollout steps.

### DIFF
--- a/plans/15-benchmark-series-identity.md
+++ b/plans/15-benchmark-series-identity.md
@@ -1,0 +1,211 @@
+# Plan 15 — Stable Benchmark Series Identity
+
+**Goal:** Let a benchmark's `over_time` history survive renames, and make every
+identity change either *adopted* or *reported* — never silently orphaned. Today
+the reset detector added in plan 09 is keyed on the two fields whose change is
+the most common cause of a reset, so it cannot see them move.
+
+**Branch name:** `feat/benchmark-series-identity`
+
+**⚠️ Read first:** plans 09 and 14 (both implemented in v1.116.0) define the
+history key and the reconciliation model this builds on. The new field must not
+enter `hash_persistent()` (`bencher/bench_cfg.py:785`) — it identifies a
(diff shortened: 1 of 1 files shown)


blooop:
Superseded by #1012, which carries this plan doc unchanged plus its implementation, tests, and validation notes.

PR #1000, opened 2026-07-29
docs: add plan 16 — inspectable, pinnable benchmark identity

blooop:
Adds `plans/16-inspectable-benchmark-identity.md`. Plan doc only — no code changes.

`BenchCfg.hash_persistent()` (`bencher/bench_cfg.py:785`) is a method on a config
object that `plot_sweep` assembles internally, so there is no way to ask "what
cache key and history key will this declaration produce?" without running the
benchmark — which is exactly what you are trying to avoid when checking that a
refactor preserved identity.

The contributing set is documented rather than exposed
(`bencher/bench_cfg.py:795-827`), and it has moved twice recently:
`CACHE_VERSION` reached 5, and `include_result_vars=False` arrived for the history
key with plans 09/14. So anything downstream that protects a long-lived trend by
asserting on "the fields that make up the key" is asserting on a *transcription*
of the rule, which keeps passing after the rule changes.

The plan adds:

- `bn.sweep_identity(...) -> SweepIdentity` — a pure function returning
  `cache_key`, `history_key`, and `summary`, implemented by delegating to the
  same conversion + hashing path `plot_sweep` uses, so there is exactly one
  implementation of the rule.
- `BenchCfg.identity()` / `BenchResult.identity` for a run that already happened.
- Exports for `config_summary` and `diff_summaries` (`bencher/history.py:135`,
  `:155`) — already the payload behind the reset warning, currently private.
- `Bench.explain_identity()`, following A2's `explain_selection()` precedent.

Purely additive: no signature changes, no hash changes. The central acceptance
test is equivalence — the pure function's keys must be byte-identical to a real
run's across every argument form `plot_sweep` accepts, since the plan's whole
value collapses if it reimplements hashing instead of delegating.

## Summary by Sourcery

Documentation:
- Add Plan 16 design document describing a sweep identity API, identity accessors on benchmark configs/results, summary/diff helpers, and an explain_identity helper, including goals, phased rollout, tests, and compatibility considerations.

### DIFF
--- a/plans/16-inspectable-benchmark-identity.md
+++ b/plans/16-inspectable-benchmark-identity.md
@@ -1,0 +1,172 @@
+# Plan 16 — Inspectable, Pinnable Benchmark Identity
+
+**Goal:** Make a benchmark's identity a value a caller can compute, print, and
+pin in a test *without running the benchmark*. Today identity exists only as a
+private method on a `BenchCfg` that `plot_sweep()` builds internally, so the only
+way to protect a long-lived trend from an accidental reset is to reimplement the
+hashing rule downstream — against a rule that has already changed twice.
+
+**Branch name:** `feat/sweep-identity-api`
+
+**⚠️ Read first:** this plan is purely additive — one new pure function and two
+accessors, all delegating to the existing `hash_persistent()`. It must not change
+any hash. The golden hashes at `test/test_hash_persistent.py:774-801` are the
+acceptance gate.
+
+---
+
+## Problem statement (with evidence)
+
+### P1 — Identity is only reachable through a live run
+
+`BenchCfg.hash_persistent(include_repeats, include_result_vars)`
+(`bencher/bench_cfg.py:785`) is a method on a config object that
+`plot_sweep()` assembles from converted variables partway through its own body
+(`bencher/bencher.py:519`). There is no way to ask "what cache key and history
+key will this declaration produce?" before committing to a run — and for an
+expensive benchmark, running it is exactly what you are trying to avoid when
+checking that a refactor preserved identity.
+
+### P2 — The contributing set is documented, not exposed
+
+Which fields participate is stated in a docstring and a comment block
+(`bencher/bench_cfg.py:795-827`): `bench_name`, `over_time`, `repeats`, `tag`,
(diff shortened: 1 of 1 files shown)


blooop:
Addressed both points in 801a62e7.

On the file:line citations — I kept them rather than replacing them with prose references. They are the plan's evidence, and every plan in `plans/` (09 through 14) cites the same way, so stripping them would make this document weaker and inconsistent with its neighbours. What I did fix is the actual rot risk you're pointing at: a citation that names only a location. Every `file:line` in the plan now names the symbol it points at, so the referent stays findable by grep after the line moves — e.g. `bencher/bench_cfg.py:795-827` is now `BenchCfg.hash_persistent`'s docstring and its title/CACHE_VERSION comment block, the golden-hash cites name `TestGoldenBenchCfgHash` and the individual test functions, and the sweep_executor cite names `SweepExecutor.convert_vars_to_params`'s string/dict guard. The Read-first block now states that line numbers are pinned to main @ 7dad0cd4 (v1.116.0) and the symbol names are the durable reference. Auditing them against that commit turned up three that were already off: the `convert_vars_to_params` cite was one line past the `def` (822 -> 821), the title-exclusion cite landed mid-comment rather than on the NOTE block (822 -> 820-823), and the CACHE_VERSION cite started on a bare comment line and stopped short of the fold itself (824-827 -> 825-830). All corrected.

On `SweepIdentity` serialization — agreed, the Coordination note was too thin to guide an implementer, so the constraints are now stated in D1 where the type is introduced: fields are primitives and immutable containers only (str/int/bool/tuple/frozen dataclasses), never a callable, a param object, or a live worker instance; the two keys are hex digest strings; the value must survive pickle unchanged and serialize through `json.dumps(asdict(ident))` with no custom encoder; equality and hashing are value-based so it works as a dict key and compares equal across processes. The rationale is stated too, since A4 wants an identity as a storage key and a key carrying a live object cannot cross a process boundary or a cache. There is a matching acceptance-criteria bullet for the pickle and JSON round-trips (tuple-to-list the only permitted normalization) and for equal identities comparing and hashing equal.

blooop:
Superseded by #1010, which carries this plan doc unchanged plus its implementation, tests, and validation notes.

PR #1002, opened 2026-07-29
docs: add plan 18 — reusable sweep declarations

blooop:
Adds `plans/18-reusable-sweep-declarations.md`. Plan doc only — no code changes.

A sweep declaration exists only as fifteen keyword arguments to a method on a live
`Bench` (`bencher/bencher.py:271-288`), consumed into a `BenchCfg` inside
`plot_sweep`'s own body. There is no object representing "this sweep", so a
declaration cannot be stored, passed, compared, or asserted on without running the
benchmark.

The cost shows up when one measurement has to run against more than one worker — a
fast and a slow backend, mocked and live dependencies, two hardware tiers. Every
argument is retyped at each call site, nothing links them, and the sites drift:
one gains a result variable, one keeps a stale constant, and because each side
still produces a valid benchmark, nothing fails. The divergence is visible only by
reading both sites side by side.

The plan adds `bn.SweepSpec` — a frozen record of `plot_sweep`'s *declarative*
arguments only (`run_cfg`, `plot_callbacks`, `auto_plot` deliberately excluded) —
with `with_()` / `merge()` composition under one documented per-field precedence
rule, `bench.plot_sweep(spec)`, and `spec.bind(worker_cls) -> dict` so a project
can assert what each environment resolves to without constructing a bench.

`input_vars` / `result_vars` **replace** rather than append (order is hashed for
inputs, and silent appending is how duplicate result variables arise — see the
companion plan 20), with an explicit `plus_result_vars()` for the additive case.

Also documents something that already exists and is close to unadvertised: the
name-based deferred form (`bn.sweep("x", bounds=...)`,
`bencher/variables/inputs.py:753-759`, resolved at `bencher/sweep_executor.py:130-145`)
plus dict `const_vars` (`bencher/bencher.py:437-438`) already let a declaration be
written without naming any class — which is what makes reuse possible at all. Its
only signpost today is a deprecation warning telling users to stop passing
`input_vars` as a dict.

Phase 1 needs no changes to `bencher.py` at all. One **OWNER DECISION**: how `tag`
ownership splits between this and plan 13's declaration bundle.

## Summary by Sourcery

Documentation:
- Add a plan document describing SweepSpec as a reusable, declarative sweep configuration object, its composition semantics, binding to worker classes, and recommended documentation patterns for class-independent sweep declarations.

### DIFF
--- a/plans/18-reusable-sweep-declarations.md
+++ b/plans/18-reusable-sweep-declarations.md
@@ -1,0 +1,191 @@
+# Plan 18 — Reusable Sweep Declarations
+
+**Goal:** Give a sweep declaration — its input, result, and constant variables,
+its tag, its title — a value object that can be built once, inspected, varied,
+and run against more than one worker class. Today that declaration exists only as
+fifteen keyword arguments to a method on a live `Bench`, so running the same
+measurement in two environments means restating all of it at both sites, with
+nothing to detect drift between them.
+
+**Branch name:** `feat/sweep-spec`
+
+**⚠️ Read first:** plan 13 owns *run* configuration (repeats, sampling
+resolution, `show`, publishing) via `@bn.benchmark(...)`. This plan owns the
+*sweep* declaration — what is measured, not how the run is driven. Keep the two
+surfaces separate; a spec that also carries `repeats` duplicates plan 13.
+
+---
+
+## Problem statement (with evidence)
+
+### P1 — A declaration is not a value
+
+`plot_sweep()` takes fifteen keyword arguments
+(`bencher/bencher.py:271-288`) and immediately consumes them into a `BenchCfg`
+inside its own body. There is no object representing "this sweep", so a
+declaration cannot be stored in a module, passed to a function, compared against
+another, or asserted on in a test without running the benchmark.
+
+### P2 — One measurement in several environments means restating everything
+
+A benchmark that must run against more than one worker — a fast and a slow
+backend, a mocked and a live dependency, two hardware tiers, two library versions
+— has one declaration and N call sites. Every argument is retyped at each site.
+Nothing links them, so the sites drift: one gains a result variable, one keeps an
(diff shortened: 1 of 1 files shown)


blooop:
Both points addressed in 6f38925e.

Field list: D1 now has a table of all seven `SweepSpec` fields — type, accepted forms, and a note per field covering whether it enters cache identity and what `None` means (unset, which preserves `plot_sweep`'s auto-discovery for the variable fields). Followed by an explicit list of the eight `plot_sweep` arguments the spec excludes, each classified as run-time or render-time with the citation. Verifying the accepted forms against `convert_vars_to_params` found one asymmetry worth documenting: sweep-shaping keys (`values`/`bounds`/`samples`) are input-only, so `bn.sweep("y", samples=3)` as a result var raises `AttributeError` because result variables are not `SweepBase` — a spec should reject that at construction, and there is now an acceptance criterion for it.

Tag ownership: new D5 states the proposed rule in the design section instead of only sketching it in Risks. A spec's `tag` sits at plan 13's "declared" tier and beats `@bn.benchmark(tag=...)`; call-site and env still win above it. Full ordering is written out as a numbered list, and D5 notes that tier 3 over tier 4 has to be resolved before the tag reaches `BenchCfg`, since the existing plumbing concatenates `run_cfg.run_tag + tag` rather than overriding. The `OWNER DECISION` marker stays, but now points at a concrete proposal to confirm.

blooop:
Follow-up commit: anchored the last bare citation to its symbol as well — the `input_vars`-as-dict deprecation warning is now cited as `Bench.plot_sweep`, `bencher/bencher.py:406-417` (the previous 406-411 cut off mid-`warnings.warn` call).

blooop:
Superseded by #1014, which carries this plan doc unchanged plus its implementation, tests, and validation notes.

PR #1003, opened 2026-07-29
plan 19 + implementation: reject unnamed parameters at resolution time

blooop:
Plan 19 **and its implementation**. `plans/19-unnamed-parameter-detection.md` stays
as the design record; the code is the last two commits.

## The defect

`param` assigns `Parameter.name` in the metaclass of the class that *declares* the
parameter. A variable declared on a plain mixin — written to be combined with a
`ParametrizedSweep` later — never passes through that metaclass, so param's MRO scan
finds it under its attribute key with `name=None`.

`_resolve_param` looks variables up by dictionary key, so such a parameter resolved
**successfully** and the failure surfaced far from the declaration. Where it surfaced
depends on the param version, which the plan now records:

| `param` | Registers with `name=None`? | Instance get/set |
|---|---|---|
| 2.3.3 | yes | **both succeed silently** → `KeyError` for result var `'None'` in `ResultCollector.store_results` |
| 2.4.1 | yes | `ValueError: Parameter name is not set`, from inside param |

Neither names the cause, and bencher supports both.

## The fix

One invariant: a parameter found under key `k` must satisfy `param.name == k`. Violating
it raises `TypeError` (found but malformed, not missing) with a message that explains the
metaclass cause and the remedy.

The object path is checked **more weakly** — `name is not None` only, no key comparison.
`bn.box()` assigns `var.name` by hand and `bn.sweep()` returns configured copies, so a
legitimate parameter object need not correspond to any attribute on the worker. A
key-equality check there would break all of those.

## Validation

- This repo: **2076 passed, 3 skipped**; 13 new tests covering both rejection paths, the
  corrected-composition path, and a no-false-positives parametrization over `box()`, both
  `sweep()` forms, a bare param object, and a bare string.
- An external project's benchmark suite — 16 benchmarks using every one of those forms plus
  multiple-inheritance task mixins — **1132 passed, zero false positives**. That was the
  result worth having: the only real risk in this change is rejecting something valid.
- One test pins the premise itself (a plain mixin's Parameter registers with `name=None`),
  so a future param release that changes this is caught here rather than silently turning
  the guard into dead code.

## Migration note

This converts a latent misconfiguration into an immediate error. Anyone currently declaring
variables on a plain mixin was already producing unnamed dataset columns or crashing in
result storage, so no working configuration changes behaviour. On param 2.3 there may be
users whose unnamed variable never reached `result_vars` and so never crashed — they will
newly see the failure, which is why it is in the CHANGELOG-worthy category.

### DIFF
--- a/bencher/sweep_executor.py
+++ b/bencher/sweep_executor.py
@@ -19,6 +19,27 @@ from bencher.job import FutureCache
 from bencher.variables.parametrised_sweep import ParametrizedSweep
 
 
+def _unnamed_parameter_error(var_type: str, name: str, owner: str, actual) -> TypeError:
+    """The error for a Parameter that never received a name from param's metaclass.
+
+    param assigns ``Parameter.name`` in the metaclass of the class that *declares*
+    the parameter, so a parameter declared on a plain (non-Parameterized) mixin is
+    found by param's MRO scan but keeps ``name=None``.  It resolves here without
+    complaint and fails much later, where ``.name`` is used as a dataset variable
+    name or as the key into the worker's result dict -- a symptom with no path back
+    to the declaration.  Name the cause instead.
+    """
+    return TypeError(
+        f"{var_type.capitalize()} variable '{name}' on {owner} has no name "
+        f"(param.name={actual!r}). This happens when the variable is declared on a "
+        f"base class that is not a param.Parameterized subclass: param assigns a "
+        f"Parameter's name in the metaclass of the class that declares it, and a "
+        f"plain mixin has no such metaclass. Make that base subclass "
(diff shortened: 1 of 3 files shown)


blooop:
No review changes needed here — the extra commit is a citation correction found while auditing the batch for the point Sourcery raised on #1000. Three line references were off by one against `main` @ `7dad0cd4`: the input/result conversion loops in `Bench.plot_sweep` are 418-421 (422 is blank) and the `const_vars` loop is 439-443 (I had 440-444, which started on a comment inside the loop and ran one line past it). The symbols and the argument are unchanged.

PR #1004, opened 2026-07-29
docs: add plan 20 — duplicate declared variables

blooop:
Adds `plans/20-duplicate-declared-variables.md`. Plan doc only — no code changes.

Nothing checks a sweep's variable lists for duplicates —
`bencher/bencher.py:418-422` (inputs, results) and `:440-444` (consts) convert
positionally with no uniqueness check. The three variable kinds then behave three
different ways.

**Result variables — silent series split.** Verified on v1.116.0:

| Declaration | Dataset | `hash_persistent(True)` |
|---|---|---|
| `["y"]` | dims `{x: 10}`, vars `['y']` | `92bca0b5…` |
| `["y", "y"]` | dims `{x: 10}`, vars `['y']` | `eb291525…` |

Identical dataset, different key. The per-variable digests fold as a sorted
*tuple*, not a set (`bencher/bench_cfg.py:841`), while the history layer keys its
column metadata by name (a dict) — so hash and history disagree about what the
config contains. The benchmark runs, reports correct numbers, and appends to a
different trend line than it appears to belong to. This is easy to hit, since
result-variable lists are commonly assembled by concatenating overlapping groups.

**Input variables — late, unhelpful crash.** `input_vars=["x", "x"]` is accepted,
then fails from xarray: `broadcasting cannot handle duplicate dimensions on a
variable: ['x', 'x', 'repeat']`.

**Constants — three behaviors.** Dict form collapses duplicates before bencher
sees them (`bencher/bencher.py:437-438`); list-of-pairs form accepts a repeat with
two different values and lets iteration order decide.

The plan validates once, after conversion, in one helper: inputs **raise** (a
repeated dimension is never intentional and already fails); constants normalize so
both spellings behave identically, with a contradiction (same name, different
values) raising; result variables get an **OWNER DECISION** between dedupe-with-warning
(recommended — it makes the hash agree with the dataset, which is already deduped)
and raising. The dedupe option is the only phase that moves a hash, so it lands
alone.

## Summary by Sourcery

Documentation:
- Add plan document 20 describing problems, proposed design, phased implementation, tests, migration strategy, risks, and coordination for handling duplicate declared variables in sweeps.

### DIFF
--- a/plans/20-duplicate-declared-variables.md
+++ b/plans/20-duplicate-declared-variables.md
@@ -9,7 +9,7 @@ input variables, an unhelpful crash deep inside xarray.
 
 **⚠️ Read first:** the result-variable case changes a hash for configurations
 that are currently accepted, so it interacts with plan 09's reset reporting. Read
-the migration section before choosing the dedupe option.
+the migration section before landing D2.
 
 ---
 
@@ -33,10 +33,10 @@ Verified on v1.116.0 with a two-variable worker, comparing
 
 The dataset is **identical** and the key is **different**. The cause is that the
 per-variable digests are folded as a sorted *tuple*, not a set
-(`bencher/bench_cfg.py:841`), so a repeated entry appears twice in the hashed
-sequence. Meanwhile the history layer keys its column metadata by name — a dict —
-so the duplicate collapses there. Hash and history therefore disagree about what
-the configuration contains.
+(`bencher/bench_cfg.py:841`, in `BenchCfg.hash_persistent`), so a repeated entry
+appears twice in the hashed sequence. Meanwhile the history layer keys its column
+metadata by name — a dict — so the duplicate collapses there. Hash and history
+therefore disagree about what the configuration contains.
 
 The practical failure is a benchmark that runs, reports correct numbers, and
 appends to a different trend line than the one it appears to belong to. Nothing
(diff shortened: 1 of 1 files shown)


blooop:
Resolved: D2 is now a decision, not a recommendation — a duplicated result variable is deduped on resolved name, first occurrence kept, with a warning.

The reasoning is that the deduped set is already what bencher stores: `ResultCollector.setup_dataset` builds `data_vars` as a name-keyed dict (`bencher/result_collector.py:233-243`) and `data_var_columns` keys history's per-column metadata the same way (`bencher/history.py:88-103`), so the duplicate collapses in both. Only `BenchCfg.hash_persistent` folds the repeat as a second sorted-tuple entry (`bencher/bench_cfg.py:841`) and moves the key. Deduping makes identity agree with the data already being written, so it is a correctness fix rather than the lenient option; raising would reject a declaration whose meaning is unambiguous and break benchmarks that produce correct data today. Raising is kept in the doc as considered-and-rejected with that reason.

Also specified the warning text concretely (names the variable, both positions, and the fix) and removed the conditional phrasing from Phased steps, Tests, Migration, and Risks so the doc has one target throughout. D1 and D3 untouched.

blooop:
Follow-up commit: fixed the three pre-existing off-by-one citations flagged in the review response rather than leaving them for later — P1's conversion loops are 418-421 and 439-443, and P5's dict collapse is 436-437. Verified against `main` @ `7dad0cd4`; no claims changed.

blooop:
Superseded by #1011, which carries this plan doc unchanged plus its implementation, tests, and validation notes.

PR #1005, opened 2026-07-29
docs: add plan 21 — per-sample fault tolerance in sweeps

blooop:
Adds `plans/21-sample-fault-tolerance.md`. Plan doc only — no code changes.

One raising sample discards an entire sweep. The exception propagates unhandled —
`store_results` calls `job_result.result()` (`bencher/result_collector.py:334`),
which re-raises the future's exception (`bencher/job.py:119`), and nothing between
the worker and `bn.run` catches it (`bencher/bench_runner.py:276-286`). Verified on
v1.116.0: a `ParametrizedSweep` whose `__call__` raises propagates straight out of
`bn.run`, taking every already-collected sample with it — no dataset, no report,
no history entry.

The asymmetry is the argument: `Bench.optimize(catch=...)` already exists so one
bad trial does not kill a study (`bencher/bencher.py:1152`, forwarded at `:1273`,
default `()` preserving fail-fast). So bencher already holds the position that an
expensive run should tolerate an individual failure — just for one of the two ways
of driving a benchmark.

The data model is also already there: result variables have a missing-value fill
(`result_missing_fill`, `bencher/history.py:52`) and history already distinguishes
"did not exist yet" from "not recorded" via each column's birth coordinate. Nothing
produces a failed sample.

The plan adds `catch=` to the sweep path, spelled exactly as on `optimize()`, and
treats accounting as part of the same change rather than a follow-up — because
tolerance without accounting is worse than fail-fast. Today nothing counts failed
samples anywhere, so `catch=` alone would let a run in which *every* sample failed
produce an all-NaN dataset, a valid-looking report, a fresh history entry, and a
successful exit. Hence `BenchResult.failed_samples` / `n_failed`, a report section,
the count in plan 10's result JSON, and `fail_on_sample_error` (bool or a failed
fraction) which raises *after* the artifacts are written.

Gating detail: a caught sample must commit nothing to the sample cache
(`bencher/job.py:119-122`), or a transient flake becomes a permanent poisoned
entry. PR #962 established that for the optimize path; phase 2 pins it for both
executor paths and must land before `catch=` is advertised.

## Summary by Sourcery

Documentation:
- Add Plan 21 design document describing per-sample fault tolerance for sweeps, failure accounting in BenchResult and reports, cache behavior on failed samples, and phased rollout steps.

### DIFF
--- a/plans/21-sample-fault-tolerance.md
+++ b/plans/21-sample-fault-tolerance.md
@@ -1,0 +1,187 @@
+# Plan 21 — Per-Sample Fault Tolerance in Sweeps
+
+**Goal:** Stop one failing sample from discarding an entire sweep, and make
+failed samples visible in the result. `Bench.optimize()` already has the knob for
+this; `plot_sweep()` does not, so the same benchmark is fault-tolerant when driven
+by Optuna and all-or-nothing when swept.
+
+**Branch name:** `feat/sweep-catch-sample-errors`
+
+**⚠️ Read first:** the default must stay fail-fast. This plan adds an opt-in, and
+the opt-in is only safe if a caught sample is kept out of the sample cache — see
+D4 and the precedent in PR #962.
+
+---
+
+## Problem statement (with evidence)
+
+### P1 — One raising sample aborts the whole sweep
+
+The exception propagates unhandled: `store_results` calls `job_result.result()`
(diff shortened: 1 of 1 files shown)


blooop:
Superseded by #1013, which carries this plan doc unchanged plus its implementation, tests, and validation notes.

PR #1006, opened 2026-07-29
docs: index plans 15-21

blooop:
Adds the `plans/README.md` index rows for plans 15–21, plus one paragraph on where
that group came from.

Kept as a separate PR on purpose: it is the only one of the eight that touches
`plans/README.md`, so the seven plan PRs cannot conflict with each other. **Merge
this one last** — its links point at files the other seven add.

| PR | Plan |
|---|---|
| #999 | 15 — Stable benchmark series identity |
| #1000 | 16 — Inspectable, pinnable benchmark identity |
| #1001 | 17 — Single-point sweep ranges |
| #1002 | 18 — Reusable sweep declarations |
| #1003 | 19 — Reject unnamed parameters at resolution time |
| #1004 | 20 — Duplicate declared variables |
| #1005 | 21 — Per-sample fault tolerance in sweeps |

## Summary by Sourcery

Documentation:
- Extend plans/README.md with table entries and brief contextual description for plans 15–21.

### DIFF
--- a/plans/README.md
+++ b/plans/README.md
@@ -34,11 +34,24 @@ without additional context. **Read the whole plan before starting it.**
 | 12 | [Portable artifact paths & cache config](12-portable-artifact-paths-and-cache-config.md) | Low | Small–Medium | Precursor to A3/A4 |
 | 13 | [Benchmark declaration bundle & run defaults](13-benchmark-declaration-and-run-defaults.md) | Low | Medium | Coordinate with 09 |
 | 14 | [Schema-evolving over_time history](14-history-schema-reconciliation.md) | — | — | **Implemented** (design record, v1.116.0) |
+| 15 | [Stable benchmark series identity](15-benchmark-series-identity.md) | Medium | Medium | Yes — closes a hole in 09 |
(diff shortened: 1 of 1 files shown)


blooop:
Added ground rule 7 here as the batch-wide answer to the citation-rot point Sourcery raised on #1000: plans cite `file:line` plus the symbol at that location, line numbers are treated as a convenience pinned to the commit the plan was written against, and an executor who finds the evidence no longer matching the code reports it in the PR instead of working around it.

It lives in the index rather than in each plan because it is a property of every document in `plans/` — including the already-merged 09-14 — not of any one of them, and the index is the only file in this batch that all seven plan PRs can share without conflicting. The paragraph on 15-21 now also names the commit the batch was written against (`main` @ 7dad0cd4, v1.116.0), which is what the new rule asks a plan author to state.

blooop:
Updated rows 17 and 19: both plans now carry their implementation in the plan PR (#1001, #1003), each validated against this repo's suite and an external project's benchmark suite. The other five remain plan-only.

PR #1022, opened 2026-07-31
feat: reachability garbage collection for the content-addressed blob store

blooop:
Follow-up to #1021. That PR added the content-addressed blob store and honestly recorded
that it had **no garbage collection**: every distinct `ResultDataSet` payload from every run
was retained until an explicit `clear_media()`/`clear_all()`, and content addressing only
deduplicates byte-identical payloads, so a nightly sweep whose payloads drift accumulated
files without bound. This PR gives the store a lifecycle.

## The real leak

The worst case is not "the cache grows"; it is a **permanent** leak on a supported feature.
A per-variable `max_time_events` limit ages an old `over_time` cell out by overwriting it
with the missing-value sentinel, and `_null_old_entries` (`bencher/result_collector.py:74`)
deliberately does *not* delete the file — `ResultDataSet` is absent from
`_MEDIA_RESULT_TYPES` for good reason (below). So before this change every aged-out payload
was orphaned on disk forever.

## Why ownership-based cleanup could not be reused

`cleanup_job_media` and `clean_orphaned_media` work because media lives in
`cachedir/{folder}/{filename}/{job_key}/`: one directory, one owner. A blob has **no owner**
— its name is its content hash, so byte-identical payloads from different job keys, sweeps
and time events collapse onto one file. Verified empirically: a 3-event × 2-sample
`over_time` sweep where two events produce the same payload yields 6 cells but 5 blobs, with
one path appearing at two different time coordinates. Deleting "this cell's blob" would
strand the other cell's live reference. That is exactly why adding `ResultDataSet` to
`_MEDIA_RESULT_TYPES` would be the wrong fix, and why per-job pruning must keep its hands
off `blobs/`.

## The reachability model

A blob is garbage iff **no live reference anywhere names it**. `blob_reachability()` unions
every blob name referenced from:

- the **`history`** diskcache — each record's *stored* dataset, not the served projection.
  This is the case a naive implementation gets wrong: the superset dataset holds dormant and
  retired columns as ordinary data variables, and old `over_time` events whose only
  reference is at `over_time=0`. A projection or an `isel(over_time=-1)` would declare all
  of those garbage.
- the **`benchmark_inputs`** diskcache — whole pickled `BenchResult` objects (`cache_results=True`).
- **`extra_roots`** — `save_result` pickles, or directories of them.

`sample_cache` is deliberately **not** a root: it stores what the worker returned, *before*
materialization, so it cannot contain a blob path (blobs are written by
`ResultCollector._materialize_dataset_value` when the value enters the dataset). It is also
the one root whose absence is recoverable — a payload still in the sample cache
re-materializes to the *same* content-addressed path on the next hit.

Matching is on the blob **filename**, not the resolved path. A name *is* a content hash, so
it is a complete, location-independent identity; this keeps references valid when a stored
dataset carries an absolute path from a cachedir that has since been moved or copied, the
case where resolving against the current `blobs/` would silently declare everything dead.

## API

```python
bn.clean_orphaned_blobs()                                      # report only (default)
bn.clean_orphaned_blobs(dry_run=False)                         # reclaim
bn.clean_orphaned_blobs(dry_run=False, extra_roots=["runs/"])  # protect saved results
bn.clean_orphaned_blobs(dry_run=False, min_age_seconds=3600)   # grace period
bn.print_orphaned_blobs()                                      # human-readable report
bn.blob_reachability()                                         # the live set, for auditing
```

`clean_orphaned_blobs` returns `(orphan_paths, total_bytes)` and defaults `dry_run=True`,
matching `clean_orphaned_media`'s shape and convention exactly. `print_orphaned_blobs`
mirrors `print_cache_stats`. Two pixi tasks sit beside the existing cache ones:
`pixi run cache-blob-orphans` (report) and `pixi run cache-blob-gc` (reclaim).

## Deliberate decisions, including one rejected

- **`dry_run=True` is the default.** Deleting a still-referenced blob silently degrades a
  report cell to a placeholder, so the safe direction must be the one you get by accident.
- **Blobs are primary storage, not a recomputable cache** — `cache_results` and
  `cache_samples` both default to `False` (`bencher/bench_cfg.py:262,267`), and a stored
  history holds paths rather than payload copies. So there is deliberately **no size- or
  age-based eviction of referenced blobs**: nothing could restore them. This is GC, not LRU.
- **An unreadable root aborts the run.** A record that cannot be deserialized may name any
  blob, so nothing is provably garbage. `clean_orphaned_blobs` then reports and deletes
  **nothing in either mode** — which also keeps a dry run an accurate preview of the real
  one — and warns with the offending entries named so they can be removed (bencher's own
  `_load_history_record` discards unreadable history records on the next run anyway).
- **`.tmp-<uuid>` partial writes and stray files in `blobs/` are never touched**; a tmp file
  may belong to a worker writing right now.
- **Rejected: adding `ResultDataSet` to `_MEDIA_RESULT_TYPES`** so `_null_old_entries` would
  delete the aged-out file. Unsound for content-addressed storage, per the dedup evidence
  above. Reachability is the only correct model here.
- **Rejected: scanning `sample_cache`.** Its values are arbitrary worker payloads (large
  DataFrames), so a recursive scan would be expensive and unbounded, and it provably cannot
  hold a reference. Documented in `_BLOB_REFERENCE_CACHES` instead.

## The `save_result` gap, stated honestly

A result written with `bencher.render.save_result` is a pickle at a user-chosen path and
**nothing records where**, so its blob references are undiscoverable and GC can strand them
(the result still loads; its dataset cells render as placeholders). This is documented in the
`clean_orphaned_blobs` docstring, in the CHANGELOG, and in `docs/how_to_use_bencher.md`, and
`extra_roots=` is the way out for a known archive. A missing or unloadable `extra_roots`
entry aborts the scan rather than running unprotected, so a typo cannot silently strand an
archive you asked to protect. The same reasoning covers a result held only in memory by a
concurrent sweep — `min_age_seconds=` is the guard, and like `clean_orphaned_media` this is
an offline maintenance step.

## Test coverage

`test/test_cache_management.py`, reusing `_TempCacheMixin`/`_make_blob` (+29 tests, 61 in the
file). `bencher/cache_management.py` is at 94% from this file alone; the two uncovered lines
are a stat/evict race.

Reachability shapes:
- history and `benchmark_inputs` are both roots; a missing cache is empty, not unreadable
- bare strings, numpy arrays, `DataArray`s and their coords are all scanned
- ordinary media paths and non-paths are not treated as references

Survival (the cases a naive implementation gets wrong):
- a blob referenced **only by a historical `over_time` event** survives
- a blob referenced **only by a retired column** (mangled name, projected away) survives
- a blob referenced **only by a dormant column** (absent from `columns` metadata) survives
- a **deduplicated** blob survives while another cell still names it
- a reference with a **stale absolute prefix** from a moved cachedir still protects
- a blob referenced from a **cached `BenchResult`** survives

Collection and reporting:
- an unreferenced blob is collected; `freed` bytes are exact
- `dry_run=True` is the default, reports accurately and deletes nothing
- `min_age_seconds` protects a recently written blob and collects an old one
- `.tmp-*` partial writes, stray files and subdirectories are left alone
- with no reference caches at all every blob is reported garbage (documents the
  default-configuration hazard explicitly)
- `print_orphaned_blobs` dry-run / real-run / truncated (>20) / aborted output

Safe degradation:
- a corrupt (undeserializable) history record collects nothing and keeps every blob
- an unopenable and an unenumerable cache each collect nothing
- a dry run reports nothing when the scan is incomplete, matching the real run
- an undeletable blob is still reported

`save_result` handling:
- a saved result is stranded **without** `extra_roots` (the documented limitation, pinned)
- an `extra_roots` file, and a directory of pickles, both protect
- a missing or unloadable `extra_roots` entry aborts rather than running unprotected

Integration against a real cachedir (`chdir`, matching `test_history_reconciliation.py`):
- `max_time_events` aging → the aged-out blob is reclaimed, live events survive
- an aged-out cell **sharing** its blob with a live cell keeps the file

Also verified manually end-to-end: a real 3-event `over_time` sweep produces 6 blobs, all 6
reachable, 0 orphans reported.

## Docs

- `CHANGELOG.md`: new Added entry; the old "`cachedir/blobs/` has no garbage collection"
  Notes bullet is replaced, since that fact has changed.
- `plans/22-grammar-phase-1-data-model.md`: amendment 9 retires the plan's "orphan cleanup
  is A4's artifact-manifest scope" claim, which conflated per-owner reclamation (still A4)
  with reclaiming what nothing references (needs no manifest).
- `docs/how_to_use_bencher.md`: a short paragraph in the `ResultDataSet` `over_time` section,
  the only place the doc mentions blobs.

## CI

- `pixi run ci` → exit 0. ruff format clean, `ruff check` clean, pylint **10.00/10**, ty
  "All checks passed!", generate-examples clean, **2435 passed, 3 skipped**, coverage
  **90%** total (`fail_under = 85`).
- `pixi run test-split` (`BENCHER_FORCE_SPLIT_RENDER=1`) → **2435 passed, 3 skipped**.
- `pixi.lock` is not in the diff.

## Summary by Sourcery

Introduce reachability-based garbage collection for the content-addressed blob store, including APIs to inspect and reclaim unreferenced blobs and integration with existing cache management.

New Features:
- Add blob_reachability API and BlobReachability dataclass to compute the set of blob filenames still referenced from cache roots and optional external archives.
- Add clean_orphaned_blobs API to report and optionally delete unreferenced blobs in cachedir/blobs with support for dry-run mode, age filtering, and extra roots.
- Add print_orphaned_blobs helper to emit a human-readable GC report and wire it into new pixi tasks for blob orphan reporting and reclamation.

Enhancements:
- Document the blob store lifecycle and reachability GC behavior in cache_management, CHANGELOG, data model plan, and user docs.
- Export new blob GC APIs from the bencher package for direct use.
- Clarify that blob content directories are handled via reachability GC rather than per-job pruning.

Tests:
- Add extensive test coverage for blob reachability computation, orphaned blob collection behavior, safety in the presence of corrupt or unreadable caches, handling of extra roots, and interaction with max_time_events aging in real cachedirs.
- Extend existing cache management tests to cover reporting behavior for orphaned blob GC in both dry-run and real-run modes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,9 +69,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `cache_results` and `cache_samples` both default to `False`, so with the defaults nothing
   on disk references the blobs of a plain (non-`over_time`) sweep, and a stored history
   holds paths rather than payload copies. GC is therefore an offline maintenance step, as
-  `clean_orphaned_media` already is — `min_age_seconds=` adds a grace period if you must run
-  it near a live sweep — and there is deliberately no size- or age-based eviction of
-  *referenced* blobs, because nothing could restore them. An unreadable cache entry makes
+  `clean_orphaned_media` already is. `min_age_seconds=` adds a grace period, but it is a
+  mitigation, not a guarantee: it protects blobs a concurrent sweep *wrote* during the
+  window, yet a sweep that deduplicates onto an existing old blob gets no protection at any
+  grace period — a content hit skips the write and never refreshes the file's mtime
+  (pinned by `test_min_age_does_not_protect_a_new_reference_to_an_old_deduplicated_blob`).
+  Run GC only when no sweep is in flight. There is deliberately no size- or age-based
+  eviction of *referenced* blobs, because nothing could restore them. An unreadable cache
+  entry makes
   absence-of-reference unprovable, so a corrupt cache collects **nothing** in either mode
   and warns with the offending entries named.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,54 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **New gallery example `example_result_dataset_1d_over_time`** under
   `reference/meta/result_types/result_dataset/`; no existing example combined
   `ResultDataSet` with `over_time`.
+- **Garbage collection for the blob store: `bn.clean_orphaned_blobs()`.** `cachedir/blobs/`
+  previously grew until an explicit `clear_media()`/`clear_all()` wiped it whole; a nightly
+  sweep whose payloads drift accumulated a file per distinct payload forever, since content
+  addressing only deduplicates byte-identical ones. The new call reclaims exactly the files
+  nothing references any more, and `dry_run=True` is the default:
+
+  ```python
+  bn.clean_orphaned_blobs()                                    # report only
+  bn.clean_orphaned_blobs(dry_run=False)                        # reclaim
+  bn.clean_orphaned_blobs(dry_run=False, extra_roots=["runs/"])  # protect saved results
+  ```
+
+  It returns `(orphan_paths, total_bytes)` like `clean_orphaned_media`, and
+  `bn.print_orphaned_blobs()` prints the same report the way `print_cache_stats` does. Two
+  pixi tasks wrap it: `pixi run cache-blob-orphans` (report) and `pixi run cache-blob-gc`
+  (reclaim). `bn.blob_reachability()` exposes the live set on its own for auditing.
+
+  The model is **reachability, not ownership**. A blob is content-addressed, so one file may
+  back cells from many job keys, sweeps and `over_time` events at once — which is precisely
+  why `cleanup_job_media`/`clean_orphaned_media` must never touch it and why their per-job
+  model could not be reused. Instead, every blob name referenced from any dataset in the
+  `benchmark_inputs` and `history` diskcaches is collected first, and only files outside
+  that set are deleted. The `history` scan reads the *stored* record rather than the served
+  projection, so a reference held only by an old `over_time` event, or by a dormant or
+  retired column, still counts as live. Matching is on the blob's filename (which *is* its
+  content hash), so a stored absolute path from a cachedir that has since moved still
+  protects its payload. `sample_cache` is deliberately not a root: it holds what the worker
+  returned, before materialization, so it cannot name a blob — and a payload still in it
+  re-materializes to the same path on the next hit.
+
+  Two limitations are explicit rather than papered over. **Results saved outside the cache
+  are invisible**: nothing records where `bencher.render.save_result` wrote, so GC can strand
+  a saved result's payloads (it still loads; its dataset cells render as placeholders). Pass
+  the archive as `extra_roots=` to protect it. **A blob is primary storage, not a cache**:
+  `cache_results` and `cache_samples` both default to `False`, so with the defaults nothing
+  on disk references the blobs of a plain (non-`over_time`) sweep, and a stored history
+  holds paths rather than payload copies. GC is therefore an offline maintenance step, as
+  `clean_orphaned_media` already is — `min_age_seconds=` adds a grace period if you must run
+  it near a live sweep — and there is deliberately no size- or age-based eviction of
+  *referenced* blobs, because nothing could restore them. An unreadable cache entry makes
+  absence-of-reference unprovable, so a corrupt cache collects **nothing** in either mode
+  and warns with the offending entries named.
+
+  Practical note on where the garbage actually comes from: a per-variable `max_time_events`
+  limit ages an old `over_time` cell out by overwriting it with the missing-value sentinel,
+  and `_null_old_entries` deliberately does *not* delete the file (a deduplicated blob may
+  still back a live cell), so every aged-out payload was an immediate permanent leak before
+  this change.
 
 ### Changed
 - **BREAKING: a `ResultDataSet` dataset cell holds a blob path string, not an index.**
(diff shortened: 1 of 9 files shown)


blooop:
**Open design question from the adversarial review (owner decision, deliberately not changed in f2cc0e5):**

`min_age_seconds` is currently an unsound guard for one concurrent-sweep case, and this PR pins the property that makes it unsound. When a concurrent sweep's payload deduplicates onto an existing blob, `materialize_blob` takes the content-hit branch (`bencher/blob_store.py:179-184`): it skips the write and never touches the file's mtime, so the blob looks arbitrarily old to any grace period even though it just gained a new live reference. `test_blob_store_races.py::test_min_age_does_not_protect_a_new_reference_to_an_old_deduplicated_blob` demonstrates this, and `test_repeated_materialize_never_rewrites_an_existing_blob` **pins mtime stability as an invariant**.

A one-line `os.utime(blob_path)` on the dedup-hit branch would make "a new reference refreshes mtime" true, turning `min_age_seconds` into a sound guard for *both* race cases (fresh write and dedup hit). Cost: one metadata syscall per cache hit, and the mtime-stability test pin would need to flip to assert the opposite property.

Options:
1. **Status quo (this PR):** GC is documented as safe only with no sweep in flight; `min_age_seconds` is a partial mitigation. Docs now state this plainly (commit f2cc0e5).
2. **Touch mtime on dedup hit:** makes the grace period sound; requires changing the pinned invariant in `test_blob_store_races.py` and re-wording the docs back to a stronger claim.

The review recommends deciding this explicitly rather than letting the test pin decide by default. Not a merge blocker either way.

blooop:
**Resolution of the open design question above:** the owner chose option 2 — long-term correctness. Implemented in #1031: `materialize_blob` now refreshes mtime on a dedup content hit (blob mtime means *last referenced*, not *created*), making `min_age_seconds` a sound guard for both concurrent-sweep cases (fresh write and dedup reference). The mtime-stability test pins here were flipped to the new contract, with no-rewrite now asserted on the publish mechanism instead of mtime. The deletion loop was verified to already stat at delete time (immediately before each unlink), so only the irreducible stat→unlink syscall instant remains, now documented precisely.

PR #1026, opened 2026-07-31
feat: make the ty type-check gate real (plan 23 P1)

blooop:
Implements **plan 23 P1** — the enforcement floor. Follows #1025 (the 3.11 floor), which
was its prerequisite.

## Summary

Enables the 9 Tier-A `ty` rules globally, adds the relaxed/strict override structure from
plan 23 D1, and adds meta-tests so the gate cannot silently regress to a no-op again.

**The gate was worse than "21 rules ignored".** The task ran `ty check` with no `--python`,
so *every* third-party import was unresolved — `param`, `panel`, `numpy`, `xarray`. That is
why `unresolved-import` had to be ignored globally, and it silently degraded every rule
that needs a third-party type. The task now passes `--python "$CONDA_PREFIX"`, resolving
the *active* environment so `-e py311` and `-e py313` each check against their own
interpreter. Every per-rule count in plan 23 §2.1 is only reproducible with that flag.

All 17 Tier-A diagnostics are **fixed rather than suppressed**, except two scoped
`# ty: ignore` comments with inline reasons: the optional `scoop` import (its
`try/except ImportError` *is* the handling) and `signal.getsignal` (ty cannot narrow an
`IntEnum` by identity against specific members).

## Latent bugs fixed on the way

| Fix | Why it mattered |
|---|---|
| `hv_dataset: hv.Dataset = None` → `\| None` | The annotation contradicted its own default |
| `plot_callback` narrowed `Callable \| None` → `Callable` | Both `_to_panes_da` and `_to_video_panes_ds` call it unguarded, so `None` was unanswerable. All four call sites already pass a real callable |
| `extra_panels` discriminated by `isinstance(ep, Viewable)` instead of `callable()` | `callable()` cannot exclude the `Viewable` arm — a `Viewable` defining `__call__` would have been **called instead of appended** |
| `ExampleEnum.to_class` parameter widened to `ClassEnum` | It narrowed a base-class parameter — an LSP violation |
| `options(self, *_args)` → `*_args, **_kwargs` | Did not match tornado's `RequestHandler.options` |
| `agg_callable` branch tests the callable | Was re-deriving `needs_agg`; the correlation is now structural, not duplicated |
| `ResultCollector.__enter__` → `Self` | The 3.10-era workaround #1025 unblocked |

## The exhaustiveness exemplar

`_resolve_auto` now returns `ResolvedReduceType` — a `Literal` over the four non-`AUTO`
members — so `to_dataset`'s `match` is exhaustive **by construction** and its
`assert_never` arm is provably dead rather than a catch-all. Proven both ways: forcing
`invalid-return-type` on shows the narrowing genuinely holds, and deleting one arm yields

```
error[type-assertion-failure]: Argument does not have asserted type `Never`
    |  Inferred type of argument is `Literal[ReduceType.MINMAX]`
```

naming the forgotten variant.

## Plan 24 was confirmed by this PR breaking on it

Converting the catch-all to `assert_never` turned **two green tests red** with
`AssertionError: Expected code to be unreachable, but got: None`. `map_plot_panes` declares
`reduce: ReduceType | None = None` and passed that `None` into `to_dataset`, whose
annotation says `ReduceType`; the old `case _:` had been absorbing it. That is exactly plan
24 §2.1 — a *complete* match is type-clean while raising at runtime, because the subject's
type was never established.

**A behaviour discrepancy fell out and is deliberately NOT fixed here.** `reduce=None`
reaching the catch-all meant `ReduceType.NONE`, whereas `to_hv_dataset`'s own default is
`ReduceType.AUTO` — so `map_plot_panes(cb)` has always **skipped reduction rather than
averaging over repeats**. This PR normalizes `None` → `NONE` at the boundary to preserve
today's behaviour exactly; correcting the default changes rendering output and needs a
phase that can own it. (Compare `BenchResult.to()`, which handles the same sentinel
correctly by omitting the argument.)

## Scope discipline

`identity.py` was pulled back **off** the strict list rather than loosening the block,
honouring its own rule that a file is added only once it is clean. It passes
`list | dict | None` to `plot_sweep`, annotated `list[ParametrizedSweep] | None` — and the
*annotation* is the wrong one, since `convert_vars_to_params` accepts
`Parameter | str | dict | tuple`. Widening a public entry point is A5's business. Initial
strict members: `sample_order.py`, `sweep_timings.py`, `blob_store.py`.

## Test plan

- `pixi run ci` green on the default env: **2417 passed, 3 skipped, 90% coverage** (gate 85%)
- `pixi run -e py311 ty` → `All checks passed!`; `pixi run -e py313` likewise via `ci`
- New `test/test_ty_gate.py` — 14 tests, green on both py311 and py313:
  - `type-assertion-failure` and all 9 Tier-A rules asserted absent from the global ignore list
  - the `--python` flag asserted present, with a message explaining what breaks without it
  - the strict ratchet asserted non-empty
  - two subprocess probes proving ty *actually* rejects a non-exhaustive match (naming the
    missing variant) and *accepts* a complete one, so the gate has neither false negatives
    nor false positives. These use a minimal standalone config on purpose: under the repo
    config the Tier-C rules are ignored and a seeded probe passes clean — the exact trap
    that made this plan's first measurements wrong.

## Review notes

- Plan 23 gains a §10 recording all of the above, per the house convention plan 22 uses.
- The two `cast()` calls in `bench_runner.py` are honest-but-unverified by design:
  `Benchable` is discriminated by `inspect.signature` parameter count, which no checker can
  narrow, so the cast states the arm the introspection established. This is the shape plan
  24 warns about.
- Next phase is P2 (collection-path live bugs B2/B3 plus C13 executor normalization), which
  plan 24 amends with a boundary-normalization precondition.

## Summary by Sourcery

Strengthen and formalize the `ty` type-checking gate by enforcing Tier-A diagnostics, making exhaustiveness checks real, and adding meta-tests and plan documentation so regressions in type-check coverage are prevented.

New Features:
- Introduce a strict type-checking ratchet for selected modules that enables Tier-C `ty` rules as errors once files are clean.
- Add meta-tests that validate the `ty` configuration, environment resolution, and exhaustiveness behavior so the type-check gate itself is tested end-to-end.

Bug Fixes:
- Ensure `ty` runs against the active pixi environment using `--python "$CONDA_PREFIX"` so third-party imports and types are correctly resolved.
- Make match handling over `ReduceType` exhaustive and correct by using a narrowed literal return type, an `assert_never` arm, and explicit handling of the `None` sentinel.
- Align several function signatures and annotations with their actual usage and intended contracts, including `hv_dataset` optionality, `plot_callback` non-nullability, request handler `options` parameters, `ExampleEnum.to_class` input type, optuna aggregation callable semantics, and `ResultCollector.__enter__` return type.
- Fix `extra_panels` dispatch so viewable panels are appended and non-viewable callables are invoked, avoiding incorrect calls on callable `Viewable` instances.
- Handle optional `scoop` imports and `signal` handlers with scoped `ty` ignores instead of global rule suppression.

Enhancements:
- Restructure `ty` rule configuration into cost-based tiers, permanently ignored rules, relaxed paths, and strict per-file overrides to support phased tightening of type checks.
- Update plan 23 documentation with an amendments section capturing configuration changes, discovered inconsistencies, and confirmed interactions with plan 24.
- Use explicit casts in bench execution to document protocol discrimination by runtime introspection rather than relying on unchecked unions.

Tests:
- Add `test/test_ty_gate.py` to assert Tier-A and exhaustiveness rules are not globally ignored, that `ty` uses the correct environment, that the strict override block is non-empty, and that seeded exhaustive/non-exhaustive matches are treated correctly by `ty`.

### DIFF
--- a/bencher/bench_plot_server.py
+++ b/bencher/bench_plot_server.py
@@ -46,7 +46,7 @@ class _CorsStaticHandler(StaticFileHandler):
         self.set_header("Access-Control-Allow-Methods", "GET, OPTIONS")
         self.set_header("Access-Control-Allow-Headers", "*")
 
-    def options(self, *_args):
+    def options(self, *_args, **_kwargs):
         self.set_status(204)
         self.finish()
 
--- a/bencher/bench_runner.py
+++ b/bencher/bench_runner.py
@@ -7,7 +7,7 @@ import webbrowser
 from collections.abc import Callable
 from copy import deepcopy
 from datetime import datetime
-from typing import Protocol, runtime_checkable
+from typing import Protocol, cast, runtime_checkable
 
 from bencher.bench_cfg import BenchCfg, BenchRunCfg, ShowMode, normalize_show
 from bencher.bench_report import BenchReport, GithubPagesCfg, Publisher
@@ -277,13 +277,15 @@ class BenchRunner:
         if len(sig.parameters) == 2:
             # BenchableV1: takes (run_cfg, report)
             report = report or BenchReport()
-            result = bench_fn(run_cfg, report)
+            # The signature check above is the discriminator for the Benchable union; no
+            # type checker can narrow on it, so state the established arm explicitly.
+            result = cast("BenchableV1", bench_fn)(run_cfg, report)
             if getattr(result, "report", None) is None:
                 result.report = report
             return result, report
 
         # BenchableV2: takes (run_cfg)
-        result = bench_fn(run_cfg)
+        result = cast("BenchableV2", bench_fn)(run_cfg)
         result_report = getattr(result, "report", None)
 
         if report is not None:
--- a/bencher/bencher.py
+++ b/bencher/bencher.py
@@ -1673,7 +1673,7 @@ class Bench(BenchPlotServer):
         target_names,
         tag,
         agg_vars=None,
-        agg_callable=None,
+        agg_callable: Callable | None = None,
         repeats=1,
     ):
         """Return an objective function compatible with ``study.optimize()``."""
@@ -1681,7 +1681,9 @@ class Bench(BenchPlotServer):
         def objective(trial: optuna.trial.Trial):
             kwargs = {iv.name: sweep_var_to_suggest(iv, trial) for iv in input_vars}
 
-            if agg_vars or repeats > 1:
+            # Non-None exactly when `needs_agg` held at the caller (agg_vars or repeats>1),
+            # so testing it here narrows the type instead of duplicating that condition.
+            if agg_callable is not None:
                 agg_combos = list(product(*(v.values() for v in agg_vars))) if agg_vars else [()]
                 all_results = [
                     self._run_optuna_job(
--- a/bencher/class_enum.py
+++ b/bencher/class_enum.py
@@ -105,7 +105,7 @@ class ExampleEnum(ClassEnum):
     Class2 = auto()
 
     @classmethod
-    def to_class(cls, enum_val: ExampleEnum) -> BaseClass:
+    def to_class(cls, enum_val: ClassEnum) -> BaseClass:
         """Convert an ExampleEnum value to its corresponding class instance.
 
         Args:
--- a/bencher/class_enum.py
+++ b/bencher/class_enum.py
@@ -109,7 +109,7 @@ class ExampleEnum(ClassEnum):
         """Convert an ExampleEnum value to its corresponding class instance.
 
         Args:
-            enum_val (ExampleEnum): The enum value to convert
+            enum_val (ClassEnum): The enum value to convert
 
         Returns:
             BaseClass: An instance of either Class1 or Class2, depending on the enum value
--- a/bencher/job.py
+++ b/bencher/job.py
@@ -15,7 +15,9 @@ from .utils import hash_sha1
 logger = logging.getLogger(__name__)
 
 try:
-    from scoop import futures as scoop_future_executor
+    # scoop is an optional extra and is not in the default environment; the
+    # except-ImportError below is the intended handling.
+    from scoop import futures as scoop_future_executor  # ty: ignore[unresolved-import]
 except ImportError as e:
     scoop_future_executor = None
 
--- a/bencher/result_collector.py
+++ b/bencher/result_collector.py
@@ -13,7 +13,7 @@ from contextlib import suppress
 from datetime import datetime
 from itertools import product
 from pathlib import Path
-from typing import Any
+from typing import Any, Self
 
 import numpy as np
 import xarray as xr
@@ -239,9 +239,7 @@ class ResultCollector:
             self._history_cache.close()
             self._history_cache = None
 
-    # typing.Self needs python 3.11 and the package floor is 3.10, so keep the
-    # concrete return type rather than take a typing_extensions dependency.
-    def __enter__(self) -> ResultCollector:  # noqa: PYI034
+    def __enter__(self) -> Self:
         return self
 
     def __exit__(self, *exc_info) -> None:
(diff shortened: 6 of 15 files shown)


blooop:
## Adversarial review pass — one regression found and fixed

Sourcery was rate-limited on this PR, so I ran an adversarial review over the diff
instead. It found **a regression I had introduced**, plus several claims the code did not
support. All fixed in b9524867; plan 23 §10 is corrected rather than left overclaiming.

### The regression (would have silently dropped report content)

Replacing `callable(ep)` with `isinstance(ep, Viewable)` in `extra_panels` and swapping the
branches was wrong: **the two predicates are not complementary.** An object that is neither
callable nor `Viewable` — a `str`, an `hv` element, a `DataFrame` — previously reached
`append`, where `Column.append` coerces it. Under the swapped form it was *called* instead,
raising `TypeError` into the surrounding `except Exception`, so the panel vanished from the
report with only a log line:

```
extra_panels=[pn.pane.Markdown]: base=3 with_extra=4 -> APPENDED
extra_panels=[str]:              base=3 with_extra=3 -> DROPPED
extra_panels=[hv.Curve]:         base=3 with_extra=3 -> DROPPED
```

`test_extra_panels.py` only covers `pn.pane.Markdown`, so nothing caught it. Now
`callable(ep) and not isinstance(ep, Viewable)`. Worth recording: the latent bug this change
was originally justified by is **hypothetical** — no panel `Viewable` is callable.

### "Behaviour-preserving" was not true

Normalizing `reduce=None` → `ReduceType.NONE` in `map_plot_panes` changed behaviour.
`to_hv_dataset` branches on `reduce == ReduceType.NONE` *before* delegating, and `None`
deliberately did not match — so it took the generic arm where holoviews infers kdims from
the xarray variables **with their units**. Normalizing early moved it onto the arm that
passes bare strings, discarding unit metadata:

```
reduce=None      -> [('theta', 'rad'), ('repeat', 'repeats')]
ReduceType.NONE  -> [('theta', None),  ('repeat', None)]
```

The normalization now lives in `_resolve_auto` — the single funnel, reached *after* every
`== NONE` branch has been decided on the raw value. That preserves behaviour on every path
and fixes the wider problem: `to_dataset`, `to_hv_dataset`, `to_hv_type`, `to_points` and
`filter` all forward `reduce` unchanged, so fixing one caller left `reduce=None` crashing
on the others. All four entry points verified to accept `None` again.

### Two gate holes closed

**The relaxed override block was far broader than the evidence.** Measured what actually
fires in those five trees with nothing silenced: **6 diagnostics**. Five of its twelve rules
were no-ops (Tier C, already globally ignored), so its real effect was disabling seven
Tier-A rules over the largest trees in the repo — including `missing-argument` across
`bencher/example/**`, the corpus that doubles as integration tests, and *the very rule that
catches a call site omitting a newly-required argument like this phase's `plot_callback`*.
Replaced by one inline ignore plus a file-precise override for the three files that cannot
take one. Verified `missing-argument` now fires in both `test/` and `bencher/example/`.

**The meta-test could not detect the bypass it exists to prevent.** It asserted only on
`[tool.ty.rules]`; adding `"bencher/**"` to a relaxed `include` turned the whole Tier-A gate
off with the suite still green. There is now a parametrized assertion that no override block
silences a protected rule for first-party package code, verified to fail on that seeded
bypass. The probe tests also assert ty's **exit code**, so a rule demoted to warning level
cannot pass.

### One claim walked back

`to_dataset`'s match is exhaustive but **not yet a compile-time guarantee**. The proof rests
on `_resolve_auto` returning a member of `ResolvedReduceType`, and the only rule that checks
that — `invalid-return-type` — is Tier B and still globally ignored. Seeding
`return ReduceType.AUTO` passes `pixi run ty`. So for a future added member, P1 moves
"silently degrades to NONE" → "crashes, still statically undetected". P12 closes it; a
caveat comment now records this at the alias.

`pixi run ci` green: **2427 passed, 3 skipped, 90% coverage, pylint 10.00/10.**

PR #1027, opened 2026-07-31
fix: surface render failures instead of only logging them

blooop:
## Problem

`to_auto` catches every plugin/callback failure so one bad plot cannot abort a whole report. That is the right call — but the failure was recorded with `logger.exception` and nothing else. Loggers are off by default in library use, so the report gets written **missing whole plots** while every caller-visible signal still says success: no warning, a zero exit code, and an HTML file that looks complete unless you already know which plot should have been there.

I hit this downstream: a benchmark's published report lost **both** its headline scatter and its rerun recording. CI was green, the pytest warnings summary was clean, and the out-of-process renderer exited 0. The only record of the failure was a log line nobody was capturing — so from the outside it was indistinguishable from "that plot was never configured".

The report is the artifact people actually read. A silently short one is worse than a loud failure.

## Change

Every failure now leaves two marks instead of one:

- a **`RenderFailedWarning`**, so an embedding test runner (pytest's warnings summary, `-W error`) sees it without configuring logging;
- a **visible pane in the report**, naming what failed and the exception, so the gap is legible to whoever opens the HTML rather than only to whoever re-runs it.

The `logger.exception` call is kept for callers already capturing it, and the swallowing contract is unchanged: one bad plot still never aborts a whole report.

Applied to all three handlers in `to_auto` / `to_auto_plots`: plot plugins, legacy plot callbacks, and user-injected `extra_panels`.

New `bencher/results/render_failure.py` holds the shared helper and the warning class.

## Tests

- `test/test_render_failure_visibility.py` — new: a failure warns, leaves a named visible pane, does not raise, and does not cost the plots that did work; plus the helper's own contract (warns, returns a pane, still logs).
- `test/test_bench_result.py` — `test_to_auto_failing_callback_logged_not_raised` renamed to `..._surfaced_not_raised` and updated: it pinned the old "silently skipped" behaviour (`len(panes) == 1`), and now asserts the warning and the visible marker alongside the working plot. This is the one intentional behaviour change.

Full suite: **2409 passed, 3 skipped**. `ruff format`/`ruff check` clean on changed files (the 6 pre-existing `ruff check` errors on `main` are untouched); `pylint` 10.00/10 on both changed source files.

## Deliberately not in scope

`map_plot_panes` renders every `PANEL_TYPES` var (image, video, rerun, dataset) in a single loop, so one raising var still discards the panes of all the others — that is why the downstream report lost the scatter *and* the recording together, rather than just one.

Fixing that means making the pane path best-effort, which collides head-on with the existing `test_bad_column_raises_rather_than_plotting_nothing` contract in `test_xy_scatter_result.py`, `test_xy_curve_result.py` and `test_xy_distribution_results.py` — and `render_data_samples` documents that it deliberately shares that path with `PaneResult.to_panes` so a chart type "behaves identically to a declared `container=`". I tried gating isolation behind a flag and could not separate the two without weakening that contract, so it belongs in its own PR with your call on which behaviour wins. Happy to take that on if you tell me which way you want it to go.

## Summary by Sourcery

Surface plot render failures to callers and report viewers instead of silently skipping failed plots while still keeping reports best-effort.

Enhancements:
- Introduce a shared render failure helper that logs, emits a warning, and returns a visible pane describing failed plot renders.
- Update automatic plotting and extra panel handling to append visible failure markers when plugins, callbacks, or injected panels raise instead of silently omitting them.

Tests:
- Add dedicated tests verifying that render failures emit warnings, produce visible failure panes, and do not abort successful plots.
- Update the existing BenchResult callback failure test to assert the new surfaced-failure behavior while preserving the non-raising contract.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,36 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `ResultDataSet` with `over_time`.
 
 ### Changed
+- **A plot that fails to render now leaves a visible mark instead of vanishing.** Report
+  building stays best-effort — one bad plot still never aborts a whole report — but the
+  failure was previously recorded with `logger.exception` and nothing else. Loggers are off
+  by default in library use, so a report could be written *missing whole plots* while every
+  caller-visible signal still said success: no warning, a zero exit code, and an HTML file
+  that looks complete unless you already know which plot should have been there.
+
+  Each swallowed failure now leaves two marks: a `bencher.RenderFailedWarning`, so an
+  embedding test runner (pytest's warnings summary, `-W error`) sees it without configuring
+  logging; and a pane in the report naming what failed and why, so the gap is legible to
+  whoever opens the HTML. The `logger` call is kept for callers already capturing it, though
+  render-failure records now come from the `bencher.results.render_failure` logger rather
+  than from each calling module's logger — code filtering on
+  `bencher.results.bench_result` to catch plot failures should filter on the new name.
+
+  Applied to every best-effort render site: plot plugins, legacy plot callbacks, and
+  user-injected `extra_panels` in `to_auto`/`to_auto_plots`; regression overlays; and the
+  optuna analysis plots, which previously hand-rolled their own differently-formatted
+  failure pane and are now unified on the shared helper.
+
+  Reports that were silently short will now show these panes. That is the point — but a
+  caller whose test run cannot tolerate the new warning can silence that half of it (the
+  report pane is not affected):
+
+  ```python
+  import warnings
+  import bencher as bn
+  warnings.filterwarnings("ignore", category=bn.RenderFailedWarning)
+  ```
+
 - **BREAKING: a `ResultDataSet` dataset cell holds a blob path string, not an index.**
   Cells were `int` indices into `BenchResult.dataset_list`, valid only inside the process
   that produced them; they are now absolute `str` paths into the blob store, so any process
--- a/bencher/__init__.py
+++ b/bencher/__init__.py
@@ -179,6 +179,7 @@ from .regression import (
 from .results.bench_result import BenchResult
 from .results.optimize_result import OptimizeResult
 from .results.pane_result import PaneResult
+from .results.render_failure import RenderFailedWarning
 from .sample_order import SampleOrder
 from .variables.parametrised_sweep import ParametrizedSweep
 from .variables.singleton_parametrized_sweep import ParametrizedSweepSingleton
--- a/bencher/optuna_conversions.py
+++ b/bencher/optuna_conversions.py
@@ -9,6 +9,7 @@ import param
 # NOTE: `optuna.visualization` pulls in sklearn's fANOVA evaluator (~3s at
 # import). It is only needed by param_importance(), so import it lazily there.
 from bencher.bench_cfg import BenchCfg
+from bencher.results.render_failure import report_render_failure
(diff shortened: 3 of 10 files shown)


blooop:
## Review — approach is right, four gaps fixed in 356c4a98

The core judgment holds: `to_auto` should stay best-effort, and a swallowed failure needs a caller-visible signal. Warning + visible pane is the right pair of marks, and the `-W error` reachability argument is the strongest part of the case. I pushed fixes rather than leaving comments.

### 1. There were four handlers in scope, not three

The PR says "all three handlers in `to_auto` / `to_auto_plots`". `to_auto_plots` has a fourth, ten lines above the `extra_panels` handler you did fix:

```python
try:
    plot_cols.append(pn.pane.HoloViews(r.render_overlay()))
except Exception:
    logger.exception("Failed to render regression overlay for %s", r.variable)
```

Identical shape, identical consequence — a failing regression overlay still vanished from the report with no signal. Now routed through the helper, with a test that builds a real over-time result and makes `render_overlay` raise.

### 2. `logger.exception` ignores the `exc` the helper was handed

`logger.exception` derives its traceback from the ambient `sys.exc_info()`, not from the exception argument. The helper takes `exc` explicitly, so calling it after the `except` block has been left is legitimate — and there it logged:

```
ERROR:bencher.results.render_failure:Thing 'z' failed
NoneType: None
```

That's the *old* problem in a new place: the record exists but carries no diagnosis. `test_helper_still_logs` didn't catch it because it only asserted the `what` string appeared. Now `logger.error(..., exc_info=exc)`, with a test that asserts a real traceback header — I verified it fails against the old line (output above is from that negative control).

### 3. The PR added a *second* failure-pane mechanism next to one that already existed

`optuna_conversions._append_safe` / `_append_safe_sized` and `run.py` already did log-and-append-a-visible-pane, in a different format:

```python
row.append(pn.pane.Markdown(f"**Plot failed** (`{fn_name}`): {e}"))
```

So a single report could show two visual styles for the same class of event, and those three sites still lacked the warning half of the fix. Since the helper exists precisely to be shared, they now use it.

### 4. `RenderFailedWarning` wasn't public

Filtering the warning is the stated point of raising it, but callers had to reach into `bencher.results.render_failure` to name the class. Exported as `bn.RenderFailedWarning`.

### Also

- Renamed `test_failing_user_plugin_logged_not_raised` → `..._surfaced_not_raised` and `test_extra_panels_callable_error_is_caught` → `..._is_surfaced`, and strengthened both. You renamed the equivalent test in `test_bench_result.py` but these two still pinned the old contract in their names and comments ("the broken panel is omitted", "logged and skipped"). They passed either way, which is exactly why they'd have rotted quietly.
- Added the CHANGELOG entry. Worth flagging one compat detail it now documents: centralizing in the helper moves render-failure log records from each calling module's logger to `bencher.results.render_failure`, so anyone filtering on `bencher.results.bench_result` to catch plot failures needs to update. I think centralizing is the right call — one logger name for all render failures — but it deserved to be written down.

### Not changed

Your `map_plot_panes` scope note is correct and I'm not reopening it here — the collision with `test_bad_column_raises_rather_than_plotting_nothing` is real and picking a winner is your call. `bench_cfg.py:973` (Cartesian animation) is the same silent-skip shape but returns a path rather than a pane, so it needs a different fix; left alone deliberately.

### Verification

`2412 passed, 3 skipped` (baseline on this branch was 2409 — the three new tests are the overlay handler, the traceback contract, and the public export). `ruff format`/`ruff check` clean; `pylint` 10.00/10 across all tracked files. I also confirmed the whole suite emits exactly two `RenderFailedWarning`s, both from tests that deliberately induce failures — so nothing was silently failing in the suite before this.

PR #1034, opened 2026-07-31
refactor: one AggFn vocabulary; unknown aggregation raises (plan 23 P11)

blooop:
Implements plan 23 P11 (C11), folding in plan 24 A3's `agg_fn` DoD addition.

**This is a small breaking change on the plotting path** — see "Behaviour changes" below.
Both breaks are in the `### Changed` CHANGELOG entry.

## The four spellings, collapsed to one

`AggFn` (`bencher/utils.py:386`, `strenum.StrEnum` with explicit lowercase values per
plan 23 D4 — `strenum`'s `auto()` would yield `'MEAN'`, and these strings are exactly
what `BenchCfg.agg_fn` accepts, so they are pinned) is now the single definition.
Everything else derives from it:

1. **The `Literal`s** — `bench_result.py:206` (`to`), `bench_result_base.py:214`
   (`to_hv_dataset`), `:293` (`to_dataset`), `:741` (`filter`) become
   `AggFn | str [| None]`. `str` is kept in the annotation because every caller in the
   tree (and `bench_cfg.agg_fn` itself) passes a raw string; the type is made true by
   normalizing, not by narrowing the signature.
2. **`AGG_FN_MAP`** (`utils.py:435`) is now keyed by `AggFn` members.
3. **The `ObjectSelector`** (`bench_cfg.py:824`) takes `default=AggFn.MEAN.value`,
   `objects=[m.value for m in AggFn]`. The accepted strings are **byte-identical** to
   before — this is internal single-sourcing only; the knob's *home* stays where it is
   (A5's business).
4. **The if/elif ladder** (`bench_result_base.py:401`) is now
   `match normalize_agg_fn(agg_fn):` over the five members with
   `case _ as unreachable: assert_never(unreachable)`. It no longer disagrees with
   `AGG_FN_MAP` because it no longer has its own vocabulary.

`normalize_agg_fn` (`utils.py:413`) is the boundary: `None` → `AggFn.MEAN` (unchanged
default semantics), otherwise `AggFn(value)` or `ValueError` naming the value and the
full vocabulary. `optimize()` (`bencher.py:1386`) now calls it too, replacing its
hand-rolled `if agg_fn not in AGG_FN_MAP: raise` — same behaviour, one implementation.

## Boundary normalization is what licenses the exhaustive match (plan 24 A2/A3)

`BenchCfg.agg_fn` is a `param.ObjectSelector` whose `objects` are **raw strings**, so a
read of that field is `Unknown` to `ty` and a `match` on it would not be an
exhaustiveness check of anything. Constructing the enum at the boundary — `AggFn(value)`,
raising on unknown input — is the precondition that licenses `assert_never` there, not
merely tidiness (plan 24 A2). Per A3 this field is strictly worse than `executor`, whose
`objects` are already enum members: the enum must be *constructed*, never assumed to be
what the descriptor holds. `test_valid_raw_string_through_param_field_reaches_shipped_path`
asserts `type(res.bench_cfg.agg_fn) is str` to pin that rather than let it drift.

### 1. Unknown `agg_fn` raises instead of silently meaning `mean`

The ladder's terminal `else` was commented "Fall back to mean if unknown string
provided", so `agg_fn="meen"` silently produced a mean-aggregated plot — while
`optimize()` raised on the same input. They agree now.

Validation is also **unconditional**: it was previously gated on `agg_over_dims` being
non-empty, so whether a typo raised depended on the dataset's dims. That is
data-dependent validation, the shape plan 23 exists to remove, and it contradicted the
headline. Now `res.to_dataset(agg_fn="bogus")` raises too.

### 2. Uppercase is no longer accepted on the plotting path

The ladder did `(agg_fn or "mean").lower()`, so `agg_fn="MEAN"` worked in
`to`/`to_dataset`/`to_hv_dataset`/`filter` — and **nowhere else**:
`plot_sweep(agg_fn="MEAN")` and `optimize(agg_fn="MEAN")` already raised, because the
`ObjectSelector`'s objects are lowercase. The leniency was undocumented, unused anywhere
in the tree, and asymmetric across the API — a fifth partial spelling of the vocabulary
rather than a feature. Folding inside `normalize_agg_fn` would recreate the exact
divergence this PR deletes (the `ObjectSelector` would still reject `"MEAN"` while
`to_dataset` accepted it); folding everywhere needs widening the accepted set, which is
A5's call. So the raise stays, and the error message names the fix:
`Unknown agg_fn='MEAN', must be one of [...] (the vocabulary is lowercase; did you mean 'mean'?)`.

## When the raise lands, honestly

Not mid-sweep for the two config ingress points:

- **`plot_sweep(agg_fn=...)`** → `BenchCfg`'s `ObjectSelector` rejects it at config
  acceptance, before any sample is collected.
  `test_unknown_agg_fn_through_plot_sweep_raises_before_any_sampling` asserts this with a
  counting worker (plus a valid-`agg_fn` control proving the count would otherwise move),
  because asserting only `ValueError` would pass if `BenchCfg` construction moved after
  sampling.
- **`to_dataset`/`to_hv_dataset`/`filter`** → raises during report building, after
  results are collected and cached.

**There is a third ingress point, and it is worse than the other two.** A user-supplied
`plot_callbacks` entry that passes `agg_fn="MEAN"` (or a typo) now raises out through
`BenchResult.plot()` → `report.append_result` → `run_sweep` → `plot_sweep` and crashes
the sweep. `bench_result.py:279-281` invokes callbacks with **no** `try`/`except`, unlike
`to_auto`'s plugin loop, so nothing contains it. Measured: `main` returns a result where
this PR crashes.

The mitigation is real but partial: `cache_results` (`bencher.py:901`) runs *before* the
`auto_plot` block (`:910`), so with caching enabled the samples are already on disk and a
re-run replots from cache. But `cache_results` and `cache_samples` both default to
`False` (verified), so **from a default-config user's seat the run's data is lost.**
That is a genuine regression in blast radius for the uppercase case specifically, and it
is stated here rather than buried.

The missing `try`/`except` at `bench_result.py:280` is **pre-existing and not fixed
here** — the coordinator is filing it separately. It is what turns any callback error,
not just this one, into a lost run.

## Tests (`test/test_bench_result_base.py`, class `TestAggFnVocabulary`)

- `test_unknown_agg_fn_raises_instead_of_silently_meaning_mean` — the pre-fix regression:
  on `main` this returns a mean-aggregated dataset.
- `test_unknown_agg_fn_raises_without_agg_over_dims` — validation is not data-dependent
  (bare, empty-list, and nonexistent-dim cases).
- `test_unknown_agg_fn_raises_on_a_warm_cache` — the cache-key call is the *only*
  validation once `_to_dataset_cache` hits, since `to_dataset` returns before the match.
- `test_unknown_agg_fn_through_plot_sweep_raises_before_any_sampling` and
  `test_unknown_agg_fn_param_assignment_raises` — **plan 24 A3 DoD**: the raw string goes
  through the `param` field into the shipped path, not into an enum-typed function.
- `test_valid_raw_string_through_param_field_reaches_shipped_path` — a valid raw string
  stored by `param` survives the shipped read into normalize+match, and
  `_scalar_aggregate_summary()` (a real consumer of `bench_cfg.agg_fn`) renders.
- `test_each_member_aggregates_as_before` — all five members, as member *and* raw string,
  against `np.nanmean/nansum/nanmax/nanmin/nanmedian` of the unaggregated data.
- `test_uppercase_agg_fn_raises` / `test_uppercase_agg_fn_error_names_the_lowercase_spelling`
  — pin the break and the migration hint.
- `test_agg_fn_map_covers_every_member` / `test_bench_cfg_objects_are_agg_fn_values` — the
  derivation is asserted, including key/object *types* (set equality alone would pass if
  `AGG_FN_MAP` reverted to raw-string keys, since `AggFn` is a `StrEnum`).
- `optimize()`'s existing raise behaviour is unchanged: `test/test_optimize.py:255` and
  `:243` pass untouched.

## Plan 23 §9 grep check

`grep -rn 'AGG_FN_MAP\|Literal\["mean"\|"mean", "sum"' bencher/` → exactly one definition
of the vocabulary (`utils.py`'s `AggFn`), plus the derived `AGG_FN_MAP` and its consumers.

## Falsified plan claims

None. Every claim in §3 C11 and §5 P11 held, modulo line drift the brief anticipated: the
`Literal`s are at `bench_result.py:206` and `bench_result_base.py:214`/`:293`/`:741`
(plan says `:175`, `:203`/`:266`/`:712`); the `ObjectSelector` (`bench_cfg.py:820-824`)
and `AGG_FN_MAP` (`utils.py:385-391`) were exact; the ladder was at `:401-428` with the
terminal `# Fall back to mean if unknown string provided` else exactly as described. The
`.lower()` case-folding is one detail C11 does not mention — a fifth spelling, not a
contradiction.

## Notes

- No cache/hash/`CACHE_VERSION` impact: `agg_fn` is in `identity.py`'s
  `EXCLUDED_FIELDS` (`:48`), the accepted set is byte-identical, and all five
  aggregations are numerically unchanged.
- Residual literals at `bencher.py:331`/`:399`/`:1326` now read
  `agg_fn: AggFn | str = AggFn.MEAN.value`. **`.value`, not the member**: `param` stores
  what it is handed, so `AggFn.MEAN` would flip `type(bench_cfg.agg_fn)` from `str` to
  `AggFn` and break the pin above (verified). The reason is recorded in `AggFn`'s
  docstring so a future cleanup does not "simplify" it.
- `docs/how_to_use_bencher.md:482` no longer hand-enumerates the vocabulary.
- Version not bumped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,44 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `n_failed > 0`, in the same way the regression report is auto-inserted.
 
 ### Changed
+- **BREAKING (small): one `AggFn` vocabulary, and an unknown `agg_fn` now raises instead
+  of silently meaning `mean`** (plan 23 P11, C11). The aggregation-function vocabulary
+  existed in four independent spellings — the `Literal`s on `to`/`to_dataset`/
+  `to_hv_dataset`/`filter`, `AGG_FN_MAP`, `BenchCfg.agg_fn`'s `ObjectSelector`, and an
+  if/elif ladder in `to_dataset` that consulted none of the others. `AggFn`
+  (`bencher/utils.py`) is now the single definition and the other three derive from it;
+  the ladder normalizes at the boundary (`normalize_agg_fn`) and then matches
+  exhaustively under `assert_never`.
+
+  Two user-visible changes, both on the plotting path:
+
+  1. **An unrecognised `agg_fn` raises `ValueError`.** The ladder's terminal `else` was
+     commented "Fall back to mean if unknown string provided", so `agg_fn="meen"`
+     silently produced a mean-aggregated plot — while `optimize()` raised on the very
+     same input. The two agree now. The raise lands at plot/aggregation time, never
+     mid-sweep: via `plot_sweep`, `BenchCfg`'s `ObjectSelector` rejects the value before
+     any sample is collected; via `to_dataset`/`filter`, results are already collected
+     and cached. Validation is also unconditional now — previously an unknown value was
+     only checked when `agg_over_dims` was non-empty, so whether you got an error
+     depended on the data.
+
+  2. **Uppercase is no longer accepted.** The ladder did `(agg_fn or "mean").lower()`,
+     so `agg_fn="MEAN"` worked in `to`/`to_dataset`/`to_hv_dataset`/`filter` — and
+     nowhere else: `plot_sweep(agg_fn="MEAN")` and `optimize(agg_fn="MEAN")` already
+     raised, because the `ObjectSelector`'s objects are lowercase. That leniency was
+     undocumented, unused anywhere in the tree, and asymmetric across the API, i.e. a
+     fifth partial spelling of the vocabulary rather than a feature. Preserving it
+     inside `normalize_agg_fn` would have recreated exactly the divergence this change
+     deletes; honouring it everywhere would mean widening the accepted set, which is a
+     public-API decision that does not belong in an internal single-sourcing change.
+     The error message names the lowercase spelling
+     (`... (the vocabulary is lowercase; did you mean 'mean'?)`), so the fix is
+     mechanical. Lowercase the string, or pass an `AggFn` member.
+
+  No cache, hash or `CACHE_VERSION` impact: `agg_fn` feeds no persistent hash (it is in
+  `identity.py`'s `EXCLUDED_FIELDS`), the accepted string set is byte-identical, and all
+  five aggregations are numerically unchanged.
+
 - **`BenchRunCfg.executor` is normalized to an `Executors` member at the sweep boundary**
   (plan 23 P2, C13). Because `Executors` is a `StrEnum`,
   `param.Selector(objects=list(Executors))` accepts the bare string `"SERIAL"` and stores a
--- a/bencher/bench_cfg.py
+++ b/bencher/bench_cfg.py
@@ -16,6 +16,7 @@ from bencher.cache_management import CACHE_VERSION
 from bencher.job import Executors
 from bencher.results.composable_container.composable_container_base import PaneLayout
 from bencher.results.laxtex_result import to_latex
+from bencher.utils import AggFn
 from bencher.variables.results import OptDir
 from bencher.variables.sweep_base import SUBSAMPLING_DIVISIONS_SAMPLES, describe_variable, hash_sha1
 from bencher.variables.time import TimeEvent, TimeSnapshot
@@ -817,9 +818,13 @@ class BenchCfg(BenchRunCfg):
         "When set, run_sweep will automatically append CurveResult (mean +/- std) "
         "and BandResult (percentile bands) with these dims collapsed.",
     )
+    # The objects derive from AggFn (the single source of the vocabulary, plan 23
+    # C11/P11) but stay plain strings: param stores whatever the caller assigns, so
+    # readers of this field must construct the enum at the boundary via
+    # normalize_agg_fn before matching on it (plan 24 A2/A3).
     agg_fn = param.ObjectSelector(
-        default="mean",
-        objects=["mean", "sum", "max", "min", "median"],
+        default=AggFn.MEAN.value,
+        objects=[m.value for m in AggFn],
         doc="Aggregation function to use when agg_over_dims is set.",
     )
 
--- a/bencher/bencher.py
+++ b/bencher/bencher.py
@@ -43,7 +43,7 @@ from bencher.results.optimize_result import OptimizeResult
 from bencher.sample_order import SampleOrder
 from bencher.sweep_executor import SweepExecutor, validate_declared_vars, worker_kwargs_wrapper
 from bencher.sweep_timings import SweepTimings, phase_timer
-from bencher.utils import AGG_FN_MAP, normalize_agg_fn, params_to_str, resolve_aggregate
+from bencher.utils import AGG_FN_MAP, AggFn, normalize_agg_fn, params_to_str, resolve_aggregate
 from bencher.variables.inputs import IntSweep
 from bencher.variables.parametrised_sweep import ParametrizedSweep
 from bencher.variables.results import ResultHmap
@@ -328,7 +328,7 @@ class Bench(BenchPlotServer):
         relationship_cb: Callable | None = None,
         plot_callbacks: list[Callable] | bool | None = None,
         aggregate: bool | int | list[str] | None = None,
-        agg_fn: str = "mean",
+        agg_fn: AggFn | str = AggFn.MEAN.value,
     ) -> list[BenchResult]:
         """Run a sequence of benchmarks by sweeping through groups of input variables.
 
@@ -396,7 +396,7 @@ class Bench(BenchPlotServer):
         plot_callbacks: list[Callable] | bool | None = None,
         sample_order: SampleOrder = SampleOrder.INORDER,
         aggregate: bool | int | list[str] | None = None,
-        agg_fn: str = "mean",
+        agg_fn: AggFn | str = AggFn.MEAN.value,
         auto_plot: bool | None = None,
     ) -> BenchResult:
         """The all-in-one function for benchmarking and results plotting.
@@ -1323,7 +1323,7 @@ class Bench(BenchPlotServer):
         sampler: optuna.samplers.BaseSampler | None = None,
         warm_start: bool = True,
         aggregate: bool | int | list[str] | None = None,
-        agg_fn: str = "mean",
+        agg_fn: AggFn | str = AggFn.MEAN.value,
         repeats: int = 1,
         tag: str = "",
         run_cfg: BenchRunCfg | None = None,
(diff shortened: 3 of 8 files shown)


blooop:
All seven review items applied in `e99a7ddc`. Thanks for the cross-branch numeric diff and
the arm-deletion check on `assert_never` — and for talking me out of nothing, the
case-folding reasoning is better stated in your version than mine, so I lifted it into the
CHANGELOG.

**1. CHANGELOG entry** (`[Unreleased]` → `### Changed`) names both user-visible changes:
the raise, and the uppercase removal — including that the leniency was undocumented,
unused in-tree, and asymmetric (`to`/`to_dataset`/`to_hv_dataset`/`filter` folded;
`plot_sweep`/`optimize` already rejected it).

**2. Migration hint** (`utils.py:427-436`): when the input is a `str` whose `.lower()` is
in the vocabulary, the message appends
`(the vocabulary is lowercase; did you mean 'mean'?)`. Pinned by
`test_uppercase_agg_fn_error_names_the_lowercase_spelling`.

**3. Validation is now unconditional** (`bench_result_base.py:297-298`). You were right
that this contradicted the headline: `fn = normalize_agg_fn(agg_fn)` always runs, and only
the *key slot* stays `None` when there are no agg dims (so those calls keep sharing one
cache entry). `test_unknown_agg_fn_raises_without_agg_over_dims` covers bare, empty-list
and nonexistent-dim.

**4. Both call sites commented** (`bench_result_base.py:326-329` and `:411-417`), plus a
`Raises:` on `_to_dataset_cache_key` — a "build a cache key" helper that raises is
surprising enough to say so in the docstring, not just in a comment. Each says explicitly
that the other does not substitute for it: the cache-key call is the only validation on a
cache hit, the match-site call is the only thing that gives the subject a type `ty` can
check. `test_unknown_agg_fn_raises_on_a_warm_cache` pins your warm-cache finding.

**5. Both weak tests strengthened.**
(a) `isinstance(k, AggFn)` on every `AGG_FN_MAP` key, `type(o) is str` on every
`ObjectSelector` object and its default — the `StrEnum` equality hole you found meant set
equality alone would survive a revert to raw-string keys.
(b) `test_unknown_agg_fn_through_plot_sweep_raises_before_any_sampling` now uses a
`CountingWorker` and asserts `calls == 0`, **plus a valid-`agg_fn` control run** asserting
the count does move — without the control the zero could just mean the counter never
worked (bencher could have been sampling a copy of the worker; it isn't, and the control
proves it).

**6. PR body amended with your third ingress point**, stated as a real regression rather
than a footnote: a user `plot_callbacks` entry passing `agg_fn="MEAN"` now raises out
through `BenchResult.plot()` → `append_result` → `run_sweep` → `plot_sweep`, with no
`try`/`except` at `bench_result.py:279-281` to contain it. I verified your mitigation
analysis holds *and* that it doesn't save the default case: `cache_results` (`:901`) does
run before the `auto_plot` block (`:910`), but `BenchRunCfg().cache_results` and
`.cache_samples` are both `False`, so with defaults the run's data is lost. Left
`bench_result.py:280` alone for your separate issue.

**7. Commit is `refactor!`** and the body opens by saying this is breaking.

**Optional items, both done** — and your warning was exactly right. I confirmed
`BenchCfg(agg_fn=AggFn.MEDIAN).agg_fn` is `<AggFn.MEDIAN: 'median'>` (type `AggFn`), so
the three residual literals at `bencher.py:331`/`:399`/`:1326` use
`AggFn.MEAN.value` — the member would have flipped `type(bench_cfg.agg_fn)` to `AggFn` and
broken the `assertIs(type(...), str)` pin. The reason is recorded in `AggFn`'s docstring so
a future "cleanup" doesn't reintroduce it. `docs/how_to_use_bencher.md:482` no longer
hand-enumerates the vocabulary; it points at `AggFn` and states the case-sensitivity and
the raise.

`pixi run ci` green on the final tree with a private temp root
(`TMPDIR=$(mktemp -d) PYTEST_DEBUG_TEMPROOT=$TMPDIR`): **2601 passed, 3 skipped**, `ty`
"All checks passed!", pylint 10.00/10. Not merging — over to you.

blooop:
Merged `origin/main` (`1642f154`, i.e. after #1038/#1036/#1035/#1037) into this branch as
`0634ba67`. **Zero conflicts** — git auto-resolved `bencher/utils.py` and `CHANGELOG.md`
cleanly, because P3's edits to `utils.py` are in `publish_file` (~`:542`) while mine add
`AggFn`/`normalize_agg_fn`/`AGG_FN_MAP` around `:386-455`, and the CHANGELOG entries
landed in different subsections.

I did not take that on trust — verified rather than assumed:

- `git diff origin/main -- CHANGELOG.md` is **purely additive**: the only `+` lines are my
  P11 entry. All of P3's, P10's and P5's `[Unreleased]` entries are intact.
- P3's `publish_file(...) -> None` annotation **and** its docstring rewrite (the
  "does not return a URL" paragraph) are present verbatim at `utils.py:542-559`.
- `git diff origin/main` over `report_export.py`, `scorecard/model.py`,
  `video_controls.py`, `bench_report.py`, `plugins/`, `rerun_result.py`, `job.py`,
  `worker_job.py`, `result_collector.py` and `pyproject.toml` is **empty** — my branch is
  byte-identical to main for every file the four merged PRs touched.
- My `bencher.py` diff vs main is only the `agg_fn` signature/import/`optimize()` changes,
  nothing of P5's.

`pixi run ty` passes after the merge, including #1037's three newly-strict files.

## One finding for a later phase: the strict-ty ratchet

`pyproject.toml:318` says "every plan 23 phase adds the modules it reworks as part of its
Definition of Done", so I tested adding my two reworked modules to the strict block. **Both
fail Tier C on pre-existing debt, so I left the block unchanged**, per the same comment's
condition that "a file is added only once it is clean":

- `bencher/results/bench_result_base.py` — ~20 diagnostics, all pre-existing and unrelated
  to `agg_fn`: `param`-descriptor dynamism, `pn.panel`/`panel.pane` used in annotation
  position, `Hashable.endswith`, `ParametrizedSweep.direction`.
- `bencher/utils.py` — 3 pre-existing: `namedtuple` in a return annotation
  (`make_namedtuple`) and two `__name__` accesses in `callable_name`.

None are introduced by this PR (verified by adding the files and reading the diagnostics —
every one sits outside the lines P11 touched). Clearing them is Tier-B/C work that belongs
to P12, not a single-sourcing refactor. Flagging it because a strict reading of that
pyproject comment would otherwise look like an unmet DoD line here.

`pixi run ci` green on the merged tree with a private temp root
(`TMPDIR=$(mktemp -d) PYTEST_DEBUG_TEMPROOT=$TMPDIR`): **2674 passed, 3 skipped**, `ty`
"All checks passed!", pylint 10.00/10. Branch is 0 commits behind main. Still not merging.

PR #1035, opened 2026-07-31
fix: report tab pairing, capability vocabulary, honest rerun missing values (plan 23 P10)

blooop:
Implements plan 23 §5 **P10** — the three small constructive fixes C9, C10, C12
(`plans/23-constructive-data-modeling.md` §3).

> **Review round 2 applied.** Independent review returned DO-NOT-MERGE on two
> real defects, both now fixed with regression tests that fail on the previously
> pushed commit: `_log_tensor` still rendered the **`-1` sentinel** as real data
> (C12's own acceptance criterion), and `has()`'s "exhaustiveness guard" enforced
> nothing statically while its runtime form could **abort a report build**. The
> two smaller carry-overs and all four PR-body accuracy corrections are applied
> below. Details in the *Review round 2* section.

## C9 — report tabs: pair each result with its tab

`BenchReport.__init__` kept two parallel lists correlated by index —
`bench_results` (appended in `append_result`) and `pane` (appended in
`append_tab`, which skips `None` panes). `append_to_result` /
`prepend_to_result` resolved `self.bench_results.index(bench_res)` into
`self.pane[idx]`, so once a result whose `plot()` returned `None` desynced them,
an in-range `idx` silently wrote into a **different result's tab**. The
`except (ValueError, IndexError)` guard only caught the trailing out-of-range
case, never the misroute.

- `bencher/bench_report.py:152` `BenchReport._result_tabs` — one list of
  `(BenchResult, tab | None)` pairs replaces the two parallel lists, so
  misrouting is unrepresentable.
- `bencher/bench_report.py:230` `BenchReport._tab_for_result` — identity lookup
  over the pairs; `append_to_result` (`:268`) and `prepend_to_result` (`:281`)
  use it.
- `bencher/bench_report.py:243` `BenchReport._append_unattributed` — the fallback
  for a result with no tab of its own. It appends to the last tab (as before) but
  now **labels the pane with its true owner** and logs a WARNING, because a
  silently misattributed pane is worse than a missing one: a reader would
  otherwise credit a regression verdict to the sweep whose tab it happens to sit
  in. (Review carry-over 3.)
- `bencher/bench_report.py:294` `BenchReport.append_tab` now **returns** the tab
  column it created (`None` when `pane is None`), which is what makes the pairing
  possible. Additive return type; every existing caller ignores it. Note this
  also changes `append_title` (`:172`), which returns `append_tab(...)` in its
  `new_tab=True` branch and so now returns the tab instead of `None` — also a
  public-contract widening, also ignored by every in-tree caller.
- `bencher/bench_report.py:155` `BenchReport.bench_results` is now a read-only
  `tuple` property derived from the pairs, so an out-of-band `append` can no
  longer desync the pairing. Every in-tree read only iterates it (`save()`
  timings loop, `_emit_json`, `test/test_render.py:342`).

Normal-path behavior is unchanged: `append_result` still registers the result,
builds the tab from `render_from or bench_res`, and routes by identity.

## C10 — capability vocabulary raises at registration

`BenchData.has` dispatched on raw strings with a `return False` fallthrough, so a
misspelled capability in a plugin's `requires` produced a plugin that was
permanently, silently never selected — indistinguishable from one whose
capability is genuinely absent.

- `bencher/plugins/bench_data.py:13` `Capability` (strenum `StrEnum`, lowercase
  values) is now the single source of the four capability names.
- `bencher/plugins/bench_data.py:27` `to_capability` normalises
  `str | Capability` and raises `ValueError` naming the bad string and the valid
  vocabulary.
- `bencher/plugins/registry.py:66` `PluginRegistry.register` validates every
  entry of `plugin.requires` — this is where `requires` is first seen, so a typo
  raises at **registration** time (registration/config-time errors may raise per
  the owner principle) instead of at silent selection time.
- `bencher/plugins/bench_data.py:77` `BenchData.has` is a complete `match` over
  `Capability` ending in `typing.assert_never` (`:101`), so a member added
  without a branch is a **check-time** `ty` error. See *Review round 2* for the
  verification and for why this is plan 24 A1's *licensed* case.
- Docstring fix: `bench_data.py:53` misattributed `requires` to `PlotFilter`; it
  lives on the plugin protocol (`bencher/plugins/plugin.py:25`
  `PlotPlugin.requires`).
- The existing correctly-spelled sites (`bencher/plugins/builtins.py:161,174`,
  `requires=frozenset({"legacy_result"})`) and external plugins passing valid
  plain strings keep working unchanged.
- `Capability` / `to_capability` exported from `bencher.plugins`;
  `docs/plot_plugin_design.md` documents the vocabulary and the raise.
- **Two never-crash-mid-run carve-outs.** Registration-time raises are licensed;
  *selection*-time ones are not, and selection runs after the expensive sweep:
  - `registry.py:125` `_ensure_entry_points_loaded` — entry-point registration is
    *lazy*, firing on the first registry lookup, so a third-party plugin failing
    validation is skipped with a visible `log.warning` (matching the existing
    "Skipping plugin entry-point" handling) rather than aborting the run.
    Explicit `register()` / `@plot_plugin(register=True)` still raises.
  - `registry.py:258-267` `explain()` — `register()` makes an invalid capability
    unreachable here, but a plugin object could be mutated post-registration, and
    both production call sites leave `select(...)` **outside** their try/except
    (`results/bench_result.py:351-357`, `registry.py:318-327`). A bad capability
    is therefore reported as a rejection *reason* plus a warning.
- **Behavior change worth recording:** `only=` bypasses the capability gate
  (`registry.py:227-241` returns before it), so a third-party plugin with an
  out-of-vocabulary `requires` was previously reachable via `only="name"` and now
  fails to register at all. Acceptable — such a plugin could never be
  automatically selected, and the failure is now loud instead of silent — but it
  is a real narrowing of what a mis-specified plugin can do.

## C12 — rerun renderer: honest missing values

`bencher/results/rerun_result.py` rendered never-sampled points as real data and
swallowed failures at DEBUG.

- Every value read goes through `result_is_missing`
  (`bencher/variables/results.py:914`), the single missing-value oracle, in
  **all four** value-reading helpers: `_log_line_graph` (`:310`),
  `_log_bar_chart` (`:329`), `_log_tensor` (`:348`) and `_log_result_var`
  (`:374`).
- `_log_line_graph` skips a missing point (logs no tick) so the line shows a
  genuine gap; surviving ticks keep their coordinate sequence numbers.
- `_log_bar_chart`'s `float(val) if val is not None else 0.0` is replaced with
  `float("nan")` for missing entries.
- `_log_tensor`'s `np.nan_to_num(arr, nan=0.0)` is **gone**. Every cell is now
  tested with `result_is_missing` and blanked to NaN **before** the float
  coercion, so both sentinel families are handled: NaN cells stay gaps, and the
  *finite* `-1` sentinel (`ResultReference`, legacy `ResultDataSet` cells) becomes
  a gap instead of plotting as real data. `value_range` is then computed from the
  surviving finite values, so no sentinel can drag the colormap floor. An
  all-missing tensor is skipped with a WARNING rather than logged as a uniform
  zero plane.
- `_log_result_var`'s `# Default:` numeric fallthrough — which a `ResultPath`
  reached as `float("/path/to/x.csv")` — is replaced with explicit dispatch on
  `param.Number` (`ResultFloat`/`ResultBool` and any future numeric); an unmapped
  type logs a WARNING naming the var and its type and is skipped.
- All four `except (KeyError, ValueError, TypeError)` handlers now log at
  **WARNING** with the entity path as context.
- Render errors still never abort — visible degradation only, per the owner
  principle.
- Plan §9 grep check: no `nan_to_num(arr, nan=0.0)` remains in
  `rerun_result.py`.

### 1. HIGH — `_log_tensor` rendered the `-1` sentinel as real data

`_log_tensor` was the **fourth** value-reading helper and the only one not routed
through `result_is_missing`; P10 mandates *every* value read. The old
`np.isfinite` filter catches the NaN family only, and `-1` is finite, so a
`ResultReference` sweep with a failed sample produced:

```
ref tensor -> [[-1. 0. 1.] [2. 3. 4.] [5. 6. 7.]]   value_range [-1.0, 7.0]
```

This also **falsified this PR body's own earlier claim** that value_range was
computed "from the finite values only, so one missing cell no longer drags the
colormap to include a fabricated 0" — it stopped including a fabricated `0` and
started including a fabricated `-1`. Doubly so because the falsified-claims
section below correctly identifies `-1` as *the* real fabrication family, and the
first pass then fixed it in three places out of four.

Fixed by masking per cell with `result_is_missing` before coercing to float.
Two new tests (`test_minus_one_sentinel_is_a_gap_not_data`,
`test_dataset_legacy_minus_one_sentinel_is_a_gap`) both **fail on the previously
pushed commit** (verified by restoring `9840fc20`'s `rerun_result.py`) — the existing tensor tests used only ResultFloat/NaN, which is
exactly why this slipped through.

### 2. MEDIUM — `has()`'s guard enforced nothing, and could abort a report

The previous `raise AssertionError(...)  # exhaustiveness guard` claimed a
static guarantee it did not provide (adding a fifth `Capability` member left
`pixi run ty` reporting `All checks passed!`), and the runtime `AssertionError`
was an unguarded report abort reachable from `explain()`.

Now a complete `match` + `typing.assert_never`. This is plan 24 **A1's licensed
case, not its excluded one**: the subject comes from `to_capability(...)`, whose
return type is established as `Capability` — it is *not* a `param` descriptor
read. Verified by seeding a fifth member into the real file and running the
repo's own gate:

```
error[type-assertion-failure]: Argument does not have asserted type `Never`
   --> bencher/plugins/bench_data.py:102:17
    |             assert_never(unreachable)
    |                          Inferred type of argument is `Literal[Capability.SEEDED_PROBE]`
```

The member was then removed and `pixi run ty` is clean again. The failure now
lives at check time; the residual runtime `ValueError` path at the `explain()`
call site is caught there (see the C10 carve-outs).

### Carry-overs applied

3. The no-tab fallback no longer misattributes silently — see
   `_append_unattributed` under C9. (`plot_callbacks = param.List(None, ...)` at
   `bench_cfg.py:809` does make the None-plot path more reachable than the plan
   implies.)
4. All four PR-body accuracy corrections are applied (see below).

### Noted, not fixed (deliberately out of scope)

5. `_log_line_graph` / `_log_bar_chart` still coerce every *present* value, so a
   present `ResultReference` plots its storage **index** as a value
   (`Scalars(0.0), Scalars(2.0)`) — the 0-D path warns while the 1-D path
   fabricates. Pre-existing, orthogonal to missingness, and a fix belongs with
   the reference-dereferencing work rather than P10. **Follow-up.**
6. The fake `rr` omits `EncodedImage`/`AssetVideo`, so the ResultImage/ResultVideo
   branches of `_log_result_var` are untested. Their behavior is unchanged by this
   PR. The reviewer independently drove the **real** rerun SDK 0.35.0 and
   confirmed `Scalars(nan)`, `BarChart([2.0, nan])`, `BarChart([nan, nan])` and
   `Tensor(nan cell, value_range=...)` all work, so the fake is not masking an API
   mismatch.

## Tests

`test/test_bench_report.py`
- `test_none_plot_does_not_misroute_following_results` — a `None`-plot result
  followed by two real ones; content for the *second* result lands in its own
  tab. **Fails on the pre-fix file** (it landed in the third result's tab — the
  silent misroute).
- `test_none_plot_then_prepend_to_second_result` — prepending to the result after
  a `None`-plot result lands at position 0 of that result's own tab. **Fails
  pre-fix** (fell through to `append()`, landing at the end).
- `test_append_to_none_plot_result_falls_back_to_append` and
  `test_prepend_to_untracked_result_is_attributed` — the fallback warns and
  labels the pane with its owning sweep.
- `test_clear_resets_results_and_tabs`,
  `test_bench_results_is_read_only_snapshot`.
- `_FakeBenchResult` gained `bench_cfg` + `plot()` so the pre-existing routing
  tests drive the real `append_result` instead of hand-appending to the two lists.

`test/test_plugins.py`
- `test_register_typoed_capability_raises` — `requires={"legacy_resutl"}` raises
  at `register()`, naming the bad string, the plugin, and every valid capability.
- `test_register_valid_capability_strings_accepted` — external plugins passing
  valid plain strings still register.
- `test_skip_entry_point_with_bad_capability` — a lazily loaded entry-point
  plugin with a bad capability is skipped with a WARNING, not raised.
- `test_select_does_not_abort_on_post_registration_capability_mutation` — a
  plugin whose `requires` went bad after registration is rejected with a reason
  instead of killing the report build.
- `test_has_unknown_capability_raises`, `test_has_capability_accepts_enum`.

`test/test_rerun_result.py` (new; drives the module-level `_log_*` helpers with a
fake `rr` module and recording, so it needs no rerun viewer round-trip and is not
optional-dep-gated) — 11 tests, covering NaN gaps in the line/bar/tensor/scalar
paths, both `-1`-sentinel cases, the `"NAN"` string sentinel, the all-missing
tensor warning, the unmapped-type warning, and the two present-value happy paths.

**Pre-fix verification, measured (not asserted).** Restoring the original
`bench_report.py` + `rerun_result.py` from `origin/main` and running the two test
files: **14 fail** — 7 in `test_bench_report.py`, 7 in `test_rerun_result.py`.
Restoring only the *round-1 pushed commit* (`9840fc20`) and running all three:
**6 fail**, exactly the six round-2 tests — the two `-1`-sentinel tensor tests,
the three fallback-attribution tests, and the post-registration-mutation test.

`pixi run ci` passes: 2614 passed, 3 skipped, pylint 10.00/10, `ty` clean, run
with a private `TMPDIR`/`PYTEST_DEBUG_TEMPROOT` so a concurrent pytest cannot
delete the numbered temp dir out from under it.

## Falsified / narrowed plan claims

Per plans-README rule 7:

1. **C12's `float(val) if val is not None else 0.0` (`:323`) was dead code, not
   merely narrow.** The plan says it "turn[s] never-sampled points into plotted
   zeros". In fact the `else 0.0` branch appears **unreachable**: xarray coerces a
   literal `None` to `nan` when building the array, in both `float` and `object`
   dtype, so `_extract_scalar` never returns `None` and `float(nan)`/`float("NAN")`
   already yielded NaN. The real fabrication on that line was the **`-1`
   sentinel** family (`ResultReference`, pre-plan-22 `ResultDataSet` cells):
   `float(-1)` is `-1.0`, plotted as a real bar. Routing through
   `result_is_missing` fixes that and makes the NaN cases explicit rather than
   accidental. Consequently `test_missing_category_is_nan_not_zero` passes
   pre-fix.
2. **`nan_to_num(arr, nan=0.0)` (`:337`) is confirmed exactly as stated** — real,
   shipped, and now removed. The plan's framing of it as a *NaN*-only problem was
   itself incomplete: the same helper also rendered the finite `-1` sentinel as
   data, which no amount of NaN handling would have caught.
3. **C9's misroute is confirmed exactly as stated.** Correction to this PR's
   earlier claim of "2 tests fail pre-fix": the true figure is **7**
   `test_bench_report.py` tests, of which only
   `test_none_plot_does_not_misroute_following_results` and
   `test_none_plot_then_prepend_to_second_result` demonstrate the misroute itself;
   the rest exercise the new tuple API and the new fallback attribution. The
   substance was right, the count was not.
4. **C10 is confirmed as a latent footgun, not a shipped defect** — as the plan
   itself says; no misspelled `requires` exists in the tree.
5. **Plan 23's C12 wording "route every value read through `result_is_missing`"
   is load-bearing and was under-delivered in the first pass** — three of four
   helpers. Recorded here because the acceptance criterion, not the plan, was at
   fault.

## Not changed (noted, out of scope)

`bencher/bench_runner.py:259` `_merge_reports` copies `source.pane` tabs into
`target` without carrying result registrations, so before this PR the target's two
lists were *guaranteed* desynced after a merge. The pairing fix makes post-merge
routing safe (merged-in source results are simply untracked and take the fallback,
which now labels itself), but carrying the pairs across a merge would change
`save(emit_json=...)` output for merged reports, so it is left alone.

### DIFF
--- a/bencher/bench_report.py
+++ b/bencher/bench_report.py
@@ -234,27 +234,60 @@ class BenchReport(BenchPlotServer):
                 return tab
         return None
 
+    @staticmethod
+    def _result_title(bench_res: BenchResult) -> str:
+        """Best-effort human title for *bench_res*, for attributing a fallback pane."""
+        cfg = getattr(bench_res, "bench_cfg", None)
+        return getattr(cfg, "title", None) or getattr(bench_res, "title", None) or "unknown sweep"
+
+    def _append_unattributed(self, bench_res: BenchResult, pane: pn.panel) -> None:
+        """Fallback for a result with no tab of its own: append to the last tab.
+
+        The pane lands under a *different* result's heading, so it is labelled
+        with its true owner and a warning is emitted. A silently misattributed
+        pane is worse than a missing one — a reader would otherwise credit this
+        content (a regression verdict, say) to the sweep whose tab it sits in.
+        """
+        title = self._result_title(bench_res)
+        logger.warning(
+            "No tab for result %r (its plot() produced nothing, or it was never "
+            "registered via append_result); appending its content to the last tab "
+            "instead, labelled with its origin",
+            title,
+        )
+        self.append(
+            pn.pane.Markdown(
+                f"**⚠️ From `{title}`** — this sweep produced no tab of its own, "
+                "so its content appears here. It does **not** belong to the sweep "
+                "above.",
+                name=f"{title} (no tab)",
+            )
+        )
+        self.append(pane)
+
     def append_to_result(self, bench_res: BenchResult, pane: pn.panel) -> None:
         """Append *pane* to the tab that belongs to *bench_res*.
 
-        Falls back to :meth:`append` (the last tab) when the result is
-        untracked or its plot() produced no tab.
+        Falls back to the last tab when the result is untracked or its plot()
+        produced no tab; the fallback labels the pane with its true owner (see
+        :meth:`_append_unattributed`).
         """
         tab = self._tab_for_result(bench_res)
         if tab is None:
-            self.append(pane)
+            self._append_unattributed(bench_res, pane)
         else:
             tab.append(pane)
 
     def prepend_to_result(self, bench_res: BenchResult, pane: pn.panel) -> None:
         """Insert *pane* at the beginning of the tab that belongs to *bench_res*.
 
-        Falls back to :meth:`append` (the last tab) when the result is
-        untracked or its plot() produced no tab.
+        Falls back to the last tab when the result is untracked or its plot()
+        produced no tab; the fallback labels the pane with its true owner (see
+        :meth:`_append_unattributed`).
         """
         tab = self._tab_for_result(bench_res)
         if tab is None:
-            self.append(pane)
+            self._append_unattributed(bench_res, pane)
         else:
             tab.insert(0, pane)
 
--- a/bencher/bench_report.py
+++ b/bencher/bench_report.py
@@ -144,7 +144,22 @@ class BenchReport(BenchPlotServer):
         self.bench_name = bench_name
         self.pane = pn.Tabs(tabs_location="above", name=self.bench_name)
         self.last_save_ms: float = 0.0
-        self.bench_results: list[BenchResult] = []
+        # Each result is stored together with the tab that was created for it
+        # (None when its plot() produced nothing), so tab routing can never
+        # resolve into a *different* result's tab (plan 23 C9). The previous
+        # representation — two parallel lists correlated by index — silently
+        # misrouted once a None plot desynced them.
+        self._result_tabs: list[tuple[BenchResult, pn.Column | None]] = []
+
+    @property
+    def bench_results(self) -> tuple[BenchResult, ...]:
+        """The results registered via :meth:`append_result`, in registration order.
+
+        Read-only snapshot: register results through :meth:`append_result` so each
+        one is paired with its tab. (A mutable list here would make direct appends
+        a silent no-op — a tuple fails loudly instead.)
+        """
+        return tuple(res for res, _tab in self._result_tabs)
 
     def clear(self) -> None:
         """Remove all tabs and results so the report can be reused between runs.
@@ -152,7 +167,7 @@ class BenchReport(BenchPlotServer):
         Not safe to call while the report is being served to a live Panel session.
         """
         self.pane.clear()
-        self.bench_results.clear()
+        self._result_tabs.clear()
 
     def append_title(self, title: str, new_tab: bool = True):
         if new_tab:
@@ -201,7 +216,6 @@ class BenchReport(BenchPlotServer):
         return label
 
     def append_result(self, bench_res: BenchResult, render_from: BenchResult | None = None) -> None:
-        self.bench_results.append(bench_res)
         title = bench_res.bench_cfg.title
         label = self._time_event_label(bench_res)
         if label:
@@ -210,30 +224,50 @@ class BenchReport(BenchPlotServer):
         # routing (append_to_result) while building the pane from another — used
         # by the BENCHER_FORCE_SPLIT_RENDER path to render from a deserialized
         # copy without breaking routing. Defaults to bench_res (normal path).
-        self.append_tab((render_from or bench_res).plot(), title)
+        tab = self.append_tab((render_from or bench_res).plot(), title)
+        self._result_tabs.append((bench_res, tab))
+
+    def _tab_for_result(self, bench_res: BenchResult) -> pn.Column | None:
+        """The tab created for *bench_res*, or None when untracked / plot() was None."""
+        for res, tab in self._result_tabs:
+            if res is bench_res:
+                return tab
+        return None
 
     def append_to_result(self, bench_res: BenchResult, pane: pn.panel) -> None:
-        """Append *pane* to the tab that belongs to *bench_res*."""
-        try:
-            idx = self.bench_results.index(bench_res)
-            self.pane[idx].append(pane)
-        except (ValueError, IndexError):
+        """Append *pane* to the tab that belongs to *bench_res*.
+
+        Falls back to :meth:`append` (the last tab) when the result is
+        untracked or its plot() produced no tab.
+        """
+        tab = self._tab_for_result(bench_res)
+        if tab is None:
             self.append(pane)
+        else:
+            tab.append(pane)
 
     def prepend_to_result(self, bench_res: BenchResult, pane: pn.panel) -> None:
-        """Insert *pane* at the beginning of the tab that belongs to *bench_res*."""
-        try:
-            idx = self.bench_results.index(bench_res)
-            self.pane[idx].insert(0, pane)
-        except (ValueError, IndexError):
-            self.append(pane)
+        """Insert *pane* at the beginning of the tab that belongs to *bench_res*.
 
-    def append_tab(self, pane: pn.panel, name: str | None = None) -> None:
-        if pane is not None:
-            if name is None:
-                name = pane.name
-            self.pane.append(pn.Column(pane, name=name))
-            self.pane.active = len(self.pane) - 1
+        Falls back to :meth:`append` (the last tab) when the result is
+        untracked or its plot() produced no tab.
+        """
+        tab = self._tab_for_result(bench_res)
+        if tab is None:
+            self.append(pane)
+        else:
+            tab.insert(0, pane)
+
+    def append_tab(self, pane: pn.panel, name: str | None = None) -> pn.Column | None:
+        """Add *pane* as a new tab and return the created tab column (None if no pane)."""
+        if pane is None:
+            return None
+        if name is None:
+            name = pane.name
+        tab = pn.Column(pane, name=name)
+        self.pane.append(tab)
+        self.pane.active = len(self.pane) - 1
+        return tab
 
     def save_index(self, directory: str = "", filename: str = "index.html") -> Path:
         """Saves the result to index.html in the root folder so that it can be displayed by github pages.
--- a/bencher/plugins/__init__.py
+++ b/bencher/plugins/__init__.py
@@ -13,7 +13,7 @@ built-in chart types onto the plugin mechanism happens in subsequent PRs."""
 from __future__ import annotations
 
 from bencher.plotting.plot_filter import PlotFilter, VarRange
-from bencher.plugins.bench_data import BenchData, CacheHandle, RunMeta
+from bencher.plugins.bench_data import BenchData, CacheHandle, Capability, RunMeta, to_capability
 from bencher.plugins.plugin import PlotPlugin, plot_plugin
 from bencher.plugins.registry import (
     ENTRY_POINT_GROUP,
@@ -29,6 +29,7 @@ __all__ = [
     "ENTRY_POINT_GROUP",
     "BenchData",
     "CacheHandle",
+    "Capability",
     "PlotFilter",
     "PlotPlugin",
     "PluginDecision",
@@ -39,5 +40,6 @@ __all__ = [
     "get_registry",
     "plot_plugin",
     "register_plugin",
+    "to_capability",
     "unregister_plugin",
 ]
--- a/bencher/plugins/bench_data.py
+++ b/bencher/plugins/bench_data.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from dataclasses import dataclass, field, replace
 from datetime import datetime
-from typing import Any, Protocol, runtime_checkable
+from typing import Any, Protocol, assert_never, runtime_checkable
 
 import xarray as xr
 from strenum import StrEnum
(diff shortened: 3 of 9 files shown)


blooop:
## Review round 2 addressed — `6a007dc7`

Both required changes are in, plus both carry-overs and all four body corrections.
The review was right on every point; the two HIGH/MEDIUM findings were real defects,
not style disagreements. PR body updated in place.

### 1. HIGH — `_log_tensor` rendered the `-1` sentinel as real data — **fixed**

Confirmed and accepted without qualification: `_log_tensor` was the **fourth**
value-reading helper and the only one I failed to route through
`result_is_missing`, while P10 mandates *every* value read. `np.isfinite` catches
the NaN family only, and `-1` is finite.

This also falsified my own PR-body claim that `value_range` came "from the finite
values only, so one missing cell no longer drags the colormap to include a
fabricated 0" — it stopped including a fabricated `0` and started including a
fabricated `-1`. Doubly embarrassing given that my falsified-claims note had
already identified `-1` as *the* real fabrication family, and I then fixed it in
three places out of four.

`bencher/results/rerun_result.py:340-352` now masks per cell with
`result_is_missing` **before** coercing to float, so both sentinel generations
become genuine NaN gaps and `value_range` is computed from what survives.

Two new tests, both failing on `9840fc20`:
- `test_minus_one_sentinel_is_a_gap_not_data` — asserts `arr[0,0]` is NaN,
  `-1.0` is absent from the array, and `value_range == [1.0, 3.0]`.
- `test_dataset_legacy_minus_one_sentinel_is_a_gap` — the legacy
  `ResultDataSet` `-1` generation.

Your diagnosis of *why* it slipped through was exactly right: both pre-existing
tensor tests used ResultFloat/NaN only.

### 2. MEDIUM — `has()`'s guard enforced nothing, and could abort a report — **fixed**

Both halves confirmed. The `raise AssertionError(...)  # exhaustiveness guard`
comment asserted a static guarantee that did not exist, and the runtime
`AssertionError` was reachable from `explain()` with `select(...)` outside the
try/except at both production call sites — a report build dying *after* the
expensive sweep.

Now a complete `match` + `typing.assert_never` (`bench_data.py:92-101`). I agree
this is plan 24 **A1's licensed case rather than its excluded one**: the subject
comes from `to_capability(...) -> Capability`, an established type, not a `param`
descriptor read. Verified against the repo's own gate by seeding a fifth member
into the real file:

```
error[type-assertion-failure]: Argument does not have asserted type `Never`
   --> bencher/plugins/bench_data.py:102:17
    |             assert_never(unreachable)
    |                          Inferred type of argument is `Literal[Capability.SEEDED_PROBE]`
```

Member removed; `pixi run ty` clean again.

I also closed the residual raise at that same call site rather than leaving it:
`registry.py:258-267` turns a post-registration capability mutation into a
rejection *reason* plus a warning, so `explain()`/`select()` cannot propagate out
of a report build at all. New test
`test_select_does_not_abort_on_post_registration_capability_mutation`.

### 3. LOW carry-over — silent misattribution in the no-tab fallback — **fixed**

Taken, and I agree with the reasoning: a misattributed pane is worse than a
dropped one, because a reader credits result A's regression verdict to sweep B.
`bench_report.py:243` `_append_unattributed` logs a WARNING naming the owning
sweep and inserts a visible attribution note ahead of the pane. Thanks for the
`plot_callbacks = param.List(None, ...)` pointer (`bench_cfg.py:809`) — that does
make the None-plot path more reachable than the plan implies. Three tests cover
it.

### 4. PR-body accuracy — all four applied

- **(a)** Corrected, and I measured rather than re-asserting: against the original
  pre-PR files **14** tests fail (7 report + 7 rerun), not 2. Only
  `test_none_plot_does_not_misroute_following_results` and
  `test_none_plot_then_prepend_to_second_result` demonstrate the misroute itself.
  Against `9840fc20`, exactly the **6** round-2 tests fail.
- **(b)** You are right and I under-stated it — the `else 0.0` branch is **dead
  code**, not narrow code. Verified: xarray coerces a literal `None` to `nan` at
  array construction in *both* `float` and `object` dtype, so `_extract_scalar`
  can never return `None`. Rule 7 note rewritten accordingly.
- **(c)** Recorded: `only=` returns before the capability gate
  (`registry.py:227-241`), so a plugin with an out-of-vocabulary `requires` was
  previously reachable by name and now fails to register at all. Acceptable
  (it could never be auto-selected, and the failure is now loud) but a real
  narrowing, now stated.
- **(d)** Recorded: `append_title` (`:172`) returns `append_tab(...)` in its
  `new_tab=True` branch, so its public contract widened from `None` to the tab
  too — I had only mentioned `append_tab`.

### 5. Noted as follow-up, not fixed

Agreed on scope: `_log_line_graph`/`_log_bar_chart` still coerce every *present*
value, so a present `ResultReference` plots its storage index as a value while the
0-D path warns. Orthogonal to missingness, and a fix belongs with
reference dereferencing rather than P10. Recorded in the PR body as a follow-up.

Also recorded: the fake `rr` omits `EncodedImage`/`AssetVideo`, so those two
branches stay untested (behavior unchanged here). Thank you for driving the real
rerun SDK 0.35.0 — knowing `Scalars(nan)`, `BarChart([2.0, nan])`,
`BarChart([nan, nan])` and `Tensor(nan cell, value_range=...)` all work is the
piece I could not get from the fake, and it retires my main worry about this
approach.

### Verification

`TMPDIR=<private> PYTEST_DEBUG_TEMPROOT=<private> pixi run ci` → exit 0:
**2614 passed, 3 skipped**, pylint 10.00/10, `ty` clean. The private temp root
also removes the earlier cross-session `/tmp/pytest-of-*` interference.

Not merging.

PR #1036, opened 2026-07-31
fix: wire all four video buttons, truthful publish_file, method-aware scorecard verdict (plan 23 P3)

blooop:
Implements **plan 23 §5 P3** (`plans/23-constructive-data-modeling.md`) — the reporting/rendering live bugs **B1**, **B4**, **B5** from §3.

## B1 — all four video buttons now exist and do what they say

`bencher/results/video_controls.py:30-35` (pre-fix) zipped four button names against a **two**-element callback list, so `zip` truncated: "Pause Videos" was bound to `reset_vid` (which *unpauses*), and "Loop Videos"/"Reset Videos" were never constructed at all.

Fixed in `VideoControls.video_controls` (`bencher/results/video_controls.py:42`): one `list[tuple[str, Callable]]` of `(label, callback)` pairs, and the callbacks are now named methods on the class, each doing what its label says:

- `VideoControls.play_videos` — `paused = False`
- `VideoControls.pause_videos` — `paused = True` (the actual bug: pause never paused)
- `VideoControls.toggle_looping` — flips one shared `loop_enabled` flag and applies it to every pane
- `VideoControls.reset_videos` — `time = 0` then `paused = False`

Two review-driven corrections to this button set:

- **The button is labelled "Toggle Looping", not "Loop Videos".** `video_container` starts every video looping, so a button reading "Loop Videos" would turn looping *off* on its first press — the opposite of its label. Looping is also driven from a single `loop_enabled` flag rather than per-pane `not vid.loop`, because a per-pane toggle inverts each video independently and never reconverges once two panes differ (`test_toggle_looping_converges_on_mixed_state`).
- **Reset's rewind is documented as a request, not a guarantee.** Panel's client-side model (`panel/models/video.ts`) arms `_blocked` on every throttled `ontimeupdate` and its `set_time` handler clears that flag and returns **without seeking**, so a `time` write landing during playback is swallowed in the browser; bokeh also only transmits *changed* properties, so writing `0` over an existing `0` sends nothing. The docstring says so, and the test is named/commented to make clear it asserts python-side state only. Unpausing has no such guard and is reliable. This is not a regression — pre-fix `reset_vid` had the same defect and was not even wired to a button — but the original wording here overstated it. The two-write nudge (`time = 1e-6; time = 0`) was deliberately **not** shipped: whether bokeh coalesces two writes to the same property in one callback cannot be verified headlessly, so shipping it would have been another unverified claim.

Buttons are built with `pn.widgets.Button(label=...)` and `button.on_click(cb)` (panel deprecates `Widget.name` in favour of `label`).

Satisfies the §9 grep-level check "no `zip(button_names`".

**Tests** (`test/test_video_controls.py`): `test_all_four_buttons_created_with_correct_labels` asserts the exact four labels in order; one test per callback drives it directly against a recorded headless `pn.pane.Video` and asserts the state change; `test_buttons_clicks_drive_the_matching_callback` round-trips through `button.param.trigger("clicks")` so the wiring (not just the callbacks) is covered.

## B4 — `publish_file` is now honestly `-> None`

`bencher/utils.py:477` `publish_file` was declared `-> str` and documented as "Returns: str: the url of the published file", but the body ends with `git("push", ...)` and returns `None`.

**Decision: annotate `-> None`, do not synthesize a URL.** Rationale from the call sites:

- `bencher/utils_rrd.py:63` (`publish_and_view_rrd`) is the **only** call site and ignores the return value entirely — it builds the viewable URL itself via its `content_callback(remote, branch_name, file_path)` parameter.
- The docstring's example advertised a two-URL scheme (git remote URL + a provider-specific viewable-HTML URL template) that the current `remote: str, branch_name: str` signature **cannot** construct — the viewable-page URL format is provider-specific and is not an input to this function. Returning a fabricated URL would be a guess.

So the fix is the truthful annotation plus a docstring that describes what the function does, states that no URL is returned and why, and documents `remote` as the `str` it actually is (the old docstring wrongly called it a `Callable` returning a tuple — a leftover from a signature that no longer exists). No behaviour change; no call site needed updating.

## B5 — method-aware verdict (improved-vs-unchanged only)

`RegressionResult.threshold` (`bencher/regression.py:59`) means percent for `method="percentage"`, **MAD-sigma** for `"adaptive"`, an **absolute delta** for `"delta"`, and an **absolute limit** for `"absolute"` — but `_verdict` (`bencher/report_export.py:189` pre-fix) compared it against `abs(change_percent)` regardless.

**Scope, as the plan specifies:** regressions were always reported correctly — `_verdict` returns `"regressed"` from `reg["regressed"]` at the top, and `bencher/scorecard/model.py:77` returns `"regressed"` before `_verdict` is even called. The unit mismatch corrupted only the **improved-vs-unchanged** distinction for non-percentage methods, and that is all this change touches.

**Design chosen: `_verdict` takes the record `Mapping` directly** rather than growing to 8 positional parameters — `_verdict(reg: Mapping) -> str` (`bencher/report_export.py:199`). This works because the shape it needs is exactly `RegressionResult.to_dict()` (`bencher/regression.py:106`), which already serializes `method`, `current_value`, `baseline_value`, `threshold`, `direction` and the four band fields when set. Both call sites feed it that shape:

- `bencher/report_export.py:332` (`compare_results`) → `_verdict(r.to_dict())`
- `bencher/scorecard/model.py:82` (`cell_verdict`) → `_core_verdict(reg)`, where `reg` is already a `to_dict()` record parsed from `summary.json`

Per-method dispatch, in helpers so each rule reads on its own:

| method | improved iff | helper |
|---|---|---|
| `percentage` + unknown fallback (mirrors `method_cells`, `regression.py:317`) | beneficial and `abs(change_percent) >= threshold` — unchanged from before | `_percent_improved` (`:273`) |
| `delta` | beneficial and `abs(current_value - baseline_value) > threshold` (absolute units; strict `>` mirrors `detect_delta`) | `_delta_improved` (`:244`) |
| `adaptive` | beneficial **and** `current_value` outside `band_lower..band_upper`, **and** — when the dual-band percent gate is present — outside `percent_band_lower..percent_band_upper` too, mirroring `detect_adaptive`'s AND gate (`regression.py:1172`). Abstains to `"unchanged"` when the bands are absent from the record. | `_adaptive_improved` (`:255`) |
| `absolute` | never — a fixed limit has no baseline to improve against (`change_percent` is NaN→`None`); now explicit rather than incidental | inline (`:203`) |

`_finite_value` (`bencher/report_export.py:190`) coerces each field to a finite float or `None`, so a missing/NaN/serialized-`None` field abstains instead of raising. Review-driven additions here:

- **`OverflowError` is now caught** — a reachable crash out of the scorecard render, which the owner never-crash principle forbids. JSON integers are arbitrary-precision, and `float(10**400)` raises `OverflowError`, which `except (TypeError, ValueError)` does not cover; a foreign or hand-edited `*.summary.json` carrying an oversized int for `current_value`/`threshold` therefore aborted the page build. Closed in the reader and in `regression._finite_or_none` (`bencher/regression.py:926`), the writer side, where the same gap made `to_dict` raise instead of emitting `null`.
- **The adaptive percent band is sorted before the membership test.** `detect_adaptive` derives it as `baseline*(1 ± pct/100)`, which **inverts** the endpoints for a negative baseline (`baseline=-100, pct=5` → `(-95.0, -105.0)`), so `pct_lower <= current <= pct_upper` could never be true and the dual-band AND gate was silently a no-op. Measured before the fix: a 0.5% beneficial move outside the MAD band but inside the 5% percent band returned `"improved"` for a negative baseline and `"passed"` for its positive mirror. The detector itself is correct because it gates on `_safe_change_percent`, which divides by `abs(baseline)` — so the pre-fix code was not the exact mirror its docstring claimed. Only the percent band is affected; the MAD band is always correctly ordered.
- `_finite_value` is kept separate from `_finite_or_none` on purpose (rather than collapsing them): the former is the **tolerant reader** for records bencher may not have written, the latter the **strict writer** for values handed to `to_dict`, where a non-numeric input is a construction-time programming error. Both are now documented as such, and the `Mapping[str, object]` / `value: object` annotations are filled in for the ty hardening this plan is doing.

**The hardcoded `regressed=False` at `scorecard/model.py:82` was correct — verified, not changed.** `cell_verdict` returns `"regressed"` at `:77` from `bool(reg.get("regressed"))` before reaching the `_verdict` call, so past that guard `reg["regressed"]` is always falsy and `_verdict` can only decide improved-vs-unchanged. Passing the record straight through now makes the flag redundant (it is read from the record), which removes the duplicated-truth smell entirely; a comment at `:81-82` records why the core verdict can never return `"regressed"` there.

**Tests:**

- `test/test_report_export.py::TestVerdictMethodUnits` — driven by **real detector output**, not hand-built dicts: `detect_delta` proves a -30% / 3-absolute-unit move stays `"unchanged"` under `max_delta=5` while a -0.6% / 6-unit move is `"improved"`; `detect_adaptive` proves a -1% move on a near-noiseless history is `"improved"` even though `abs(change_percent) < threshold` (the sigma number), and that a beneficial -5% move **inside** a wide MAD band is **not** misclassified `"improved"` just because `abs(change_percent) >= 3.5`; `detect_absolute` abstains.
- `test/test_scorecard.py::TestCellVerdictMethodUnits` — the same distinctions at the `cell_verdict` layer, plus the dual-band gate (outside MAD but inside the percent band → `"passed"`; outside both → `"improved"`), missing-bands abstention, and unknown-method → percentage fallback. `_reg` (`test/test_scorecard.py:55`) gained `method=` and `**extra` so band fields can be supplied; every existing percentage-method test keeps passing unchanged.

All new tests were confirmed to fail on the pre-fix code (13 failures with `bencher/` reverted). **To be precise about how much of that is behavioural evidence:** 7 of the 13 are vacuous — the 3 `TestVerdictMethodUnits` cases fail with `TypeError: _verdict() missing 3 required positional arguments` (the signature changed, so the old function cannot be called at all) and the 4 video-callback cases fail with `AttributeError` (the methods did not exist). The genuinely behavioural evidence is `test_all_four_buttons_created_with_correct_labels`, `test_buttons_clicks_drive_the_matching_callback`, and all 4 `TestCellVerdictMethodUnits` cases, which go through `cell_verdict` (unchanged signature) and therefore demonstrate real improved→passed / passed→improved verdict diffs. The plan §9 DoD is satisfied; the wording above simply should not be read as 13 independent behavioural findings.

## Historical-record impact (no schema bump)

`SCHEMA_VERSION` is deliberately **not** bumped: the JSON structure is byte-identical, and nothing reads the field — `bencher/scorecard/discover.py:56-64` has no version gate. But that also means **every `*.summary.json` already on disk is re-read with the corrected rules on the next scorecard build**, so `adaptive`, `delta`, and `absolute` cells can move between `improved` and `passed` with no file changing and no benchmark re-running (`regressed` and `trend` cells are unaffected). `test_adaptive_missing_bands_abstains` fails pre-fix precisely because of this. Recorded in `CHANGELOG.md` so a maintainer seeing recoloured cells has something to point at, and `docs/scorecard.md` now states the policy behind it: `schema_version` is a **structural** version covering the JSON's shape, not the meaning of its values, so verdict semantics may be corrected within a version.

## Missing B4 regression test (plan §9 DoD)

Added `test/test_utils.py::TestPublishFileContract`. `publish_file` is `# pragma: no cover` and nothing else references it, so the contract is asserted without executing git: `inspect.signature(...).return_annotation` is `"None"` (a string, since `bencher/utils.py` uses `from __future__ import annotations`), plus an AST walk asserting no `return <expr>` exists in the body. Verified failing against the pre-fix `-> str` at `398bdd96`.

## Falsified plan claims

None. Every B1/B4/B5 claim in §3 held exactly as written, including the precise B5 scoping (regressions unaffected; only improved-vs-unchanged corrupted) and the note that `scorecard/model.py:82`'s `regressed=False` is likely correct.

## Verification

`pixi run ci` passes: format clean, ruff clean, pylint 10.00/10, `ty` clean, 2615 passed / 3 skipped, coverage 90%.

One environment note, not a code issue: an initial `pixi run ci` reported 88 setup **errors**, all `FileNotFoundError: /tmp/pytest-of-kinisi-ci/pytest-1` — a concurrent agent sharing this user's pytest tmp base had garbage-collected the numbered basetemp mid-run. Re-running with a private `TMPDIR` gives a fully green suite; the affected tests also pass in isolation. No version bump in `pyproject.toml`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,57 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- **Scorecard and A/B verdicts now measure an improvement in the units the detector
+  actually used** (plan 23 P3, B5). `RegressionResult.threshold` means a percent for
+  `regression_method="percentage"`, but **MAD-sigma** for `"adaptive"`, an **absolute
+  delta** for `"delta"`, and an **absolute limit** for `"absolute"` — and the verdict
+  helper compared it against `abs(change_percent)` regardless. Regressions were never
+  affected (both call sites resolve `regressed` from the detector before the comparison
+  is reached); what was corrupted is the **improved-vs-unchanged** split for the three
+  non-percentage methods: a tiny but real improvement on a quiet metric read as
+  "unchanged", and a beneficial move well inside the acceptance band read as "improved"
+  because its percent number happened to exceed a sigma count. Each method is now judged
+  on its own terms — `delta` on `|current - baseline|`, `adaptive` against the MAD
+  acceptance band (plus the percent band when the dual-band gate is configured),
+  `absolute` abstaining because a fixed limit has no baseline to improve on — and a
+  record missing the fields a method needs abstains instead of guessing.
+
+  **This recolours existing scorecards.** `schema_version` is a *structural* version and
+  the discovery pass has no version gate, so every `*.summary.json` already on disk is
+  re-read with the corrected rules on the next scorecard build: cells from `adaptive`,
+  `delta`, and `absolute` gates can move between `improved` and `passed` with no file
+  changing and no benchmark re-running. `regressed` and `trend` cells are unaffected. See
+  `docs/scorecard.md` for the versioning policy this follows.
+
+- **All four video-control buttons exist and each does what its label says** (plan 23 P3,
+  B1). Four button labels were zipped against a two-element callback list, so `zip`
+  truncated the row: only two buttons were ever built, "Pause Videos" was wired to the
+  callback that *unpauses*, and the Loop and Reset buttons did not exist. All four are now
+  built from a single list of `(label, callback)` pairs. Looping is driven by one shared
+  flag so a click moves every video pane to the same state instead of inverting each
+  independently, and the button is labelled "Toggle Looping" because videos start out
+  looping — a button reading "Loop Videos" would have turned looping *off* when first
+  pressed. Note that Reset's rewind is a request panel's client-side model can drop while
+  a video is playing (its `set_time` handler returns without seeking when a recent
+  `timeupdate` armed its internal guard); the docstring records this rather than promising
+  a rewind bencher cannot deliver.
+
+- **`publish_file` no longer claims to return a URL it never returned** (plan 23 P3, B4).
+  It was annotated `-> str` and documented as returning the published file's URL, but the
+  body ends at the `git push` and falls off the end returning `None` — so a caller
+  following the signature got `None`. The annotation and docstring now state that, and
+  document `remote` as the string it is (it was described as a callable returning a pair
+  of URLs, from a signature that no longer exists). Behaviour is unchanged: the viewable
+  URL is provider-specific and not derivable from the arguments, so callers still build it
+  themselves, as `publish_and_view_rrd` already did.
+
+- **A malformed or foreign `*.summary.json` no longer aborts a scorecard build.** JSON
+  integers are arbitrary-precision, so an oversized integer where a metric value was
+  expected raised `OverflowError` (not caught by the `TypeError`/`ValueError` guard) out of
+  the verdict pass and out of the page render; it now degrades to an abstaining verdict.
+  The same gap is closed in the writer (`RegressionResult.to_dict`), which now emits `null`
+  for such a value rather than raising.
+
 - **Fail-loud, not fail-fatal: a benchmark that returns `None` no longer produces a
   silently empty sweep — and no longer aborts the run either** (plan 23 P2, B3; amended
   per owner decision before release). A worker that forgot to `return
--- a/bencher/regression.py
+++ b/bencher/regression.py
@@ -924,10 +924,18 @@ def _clean_1d(a: np.ndarray) -> np.ndarray:
 
 
 def _finite_or_none(value: float | None) -> float | None:
-    """Coerce a float to a strict-JSON-safe value: non-finite (NaN/inf) -> None."""
+    """Coerce a float to a strict-JSON-safe value: non-finite (NaN/inf) -> None.
+
+    An int too large to convert to a float (JSON integers are
+    arbitrary-precision) is not JSON-safe either, so it degrades to ``None``
+    rather than raising ``OverflowError`` out of ``to_dict``.
+    """
     if value is None:
         return None
-    value = float(value)
+    try:
+        value = float(value)
+    except OverflowError:
+        return None
     return value if np.isfinite(value) else None
 
 
--- a/bencher/report_export.py
+++ b/bencher/report_export.py
@@ -187,16 +187,26 @@ def _snapshot_ds(bench_res: BenchResult) -> xr.Dataset:
     return ds
 
 
-def _finite_value(value) -> float | None:
-    """Coerce *value* to a finite float, or None (handles None/NaN/inf/non-numbers)."""
+def _finite_value(value: object) -> float | None:
+    """Coerce *value* to a finite float, or None for anything that isn't one.
+
+    The tolerant *reader* counterpart to
+    :func:`~bencher.regression._finite_or_none` (the strict *writer* used by
+    ``to_dict``): this one is fed records parsed from ``*.summary.json`` files
+    that bencher may not have written, so every non-numeric, non-finite, or
+    unrepresentable input degrades to ``None`` (which the verdict branches treat
+    as "abstain") rather than aborting a scorecard build. ``OverflowError``
+    matters in particular because JSON integers are arbitrary-precision, so a
+    hand-edited or foreign summary can carry an int too large for a float.
+    """
     try:
-        value = float(value)
-    except (TypeError, ValueError):
+        number = float(value)  # type: ignore[arg-type]
+    except (TypeError, ValueError, OverflowError):
         return None
-    return value if np.isfinite(value) else None
+    return number if np.isfinite(number) else None
 
 
-def _verdict(reg: Mapping) -> str:
+def _verdict(reg: Mapping[str, object]) -> str:
     """Classify a metric movement as improved / regressed / unchanged.
 
     *reg* follows the :meth:`~bencher.regression.RegressionResult.to_dict`
@@ -234,25 +244,30 @@ def _verdict(reg: Mapping) -> str:
     return "improved" if improved else "unchanged"
 
 
-def _beneficial(delta: float, direction) -> bool:
+def _beneficial(delta: float, direction: object) -> bool:
     """True when a signed movement is in the variable's beneficial direction."""
     return (direction == OptDir.minimize.value and delta < 0) or (
         direction == OptDir.maximize.value and delta > 0
     )
 
 
-def _delta_improved(reg: Mapping) -> bool:
-    """delta method: gate the improvement on |current - baseline| in absolute units."""
+def _delta_improved(reg: Mapping[str, object]) -> bool:
+    """delta method: gate the improvement on |current - baseline| in absolute units.
+
+    Strict ``>`` to mirror :func:`~bencher.regression.detect_delta`'s
+    ``delta > max_delta`` exactly (contrast :func:`_percent_improved`, which keeps
+    the pre-existing inclusive convention).
+    """
     current = _finite_value(reg.get("current_value"))
     baseline = _finite_value(reg.get("baseline_value"))
     threshold = _finite_value(reg.get("threshold"))
     if current is None or baseline is None or threshold is None:
         return False
     delta = current - baseline
-    return _beneficial(delta, reg.get("direction")) and abs(delta) >= threshold
+    return _beneficial(delta, reg.get("direction")) and abs(delta) > threshold
 
 
-def _adaptive_improved(reg: Mapping) -> bool:
+def _adaptive_improved(reg: Mapping[str, object]) -> bool:
     """adaptive method: improved iff outside the MAD band (and any percent band)."""
     current = _finite_value(reg.get("current_value"))
     baseline = _finite_value(reg.get("baseline_value"))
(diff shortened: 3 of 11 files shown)


blooop:
## Review nits addressed (`1f901b1b`)

Thanks — all seven actionable items applied, and both endorsed judgement calls (`-> None` on `publish_file`, no `SCHEMA_VERSION` bump) left as they were. Every claim in the review reproduced exactly as described; measurements below.

### 1. `OverflowError` out of the scorecard render — fixed
Confirmed reachable: `float(10**400)` raises `OverflowError`, which `except (TypeError, ValueError)` does not cover, and JSON integers are arbitrary-precision. Added to `_finite_value` (`bencher/report_export.py:190`) and to `regression._finite_or_none` (`bencher/regression.py:926`) — the writer side had the identical gap, where it made `to_dict` raise instead of emitting `null`. Tests: `TestCellVerdictHostileRecords::test_oversized_json_integer_does_not_raise` (+ a non-numeric case) and `TestRegressionDataclassToDict::test_oversized_int_becomes_none`. The docstring's "handles non-numbers" claim is now true.

### 2. Dual-band percent gate was a no-op for negative baselines — fixed
Reproduced the exact asymmetry you measured:

```
baseline=-100, pct=5%  ->  percent band (-95.0, -105.0)   # endpoints inverted
0.5% beneficial move, outside MAD band, inside percent band:
  negative baseline -> "improved"
  positive mirror   -> "unchanged"
```

Now `pct_lower, pct_upper = sorted((pct_lower, pct_upper))` before the membership test, with a comment recording why the detector is unaffected (it gates on `_safe_change_percent`, which divides by `abs(baseline)`). Test: `test_negative_baseline_percent_band_still_suppresses` asserts the negative record and its positive mirror now agree. Confirmed only the percent band needed this — the MAD band is `baseline ± mad·noise_floor` with `noise_floor > 0`, always ordered.

### 3. Reset claim softened — took (b), deliberately not (a)
Confirmed in the installed `panel/models/video.ts`: `update_time()` (`:90-101`) sets `_blocked = true` on every throttled `ontimeupdate`, and `set_time()` (`:146-153`) early-returns clearing `_blocked` **without seeking**. `set_paused()` (`:124-132`) has no such guard, so the unpause half is reliable.

`reset_videos`'s docstring now states the rewind is a *request*, names the upstream guard and the changed-property-only transmission, and says the python-side state is what's guaranteed. `test_reset_callback_rewinds_and_plays` → `test_reset_callback_sets_python_state_and_plays`, with a docstring saying it is not evidence of a browser rewind. PR body corrected.

I **did not** ship the two-write nudge. Whether bokeh coalesces two writes to the same property within one callback is exactly the sort of thing I can't verify headlessly, so shipping it would have replaced one unverified claim with another — which is what this review is asking me to stop doing. Flagging instead: a reliable rewind needs the upstream `set_time` guard fixed.

### 4. "Loop Videos" turned looping off — renamed + single flag
Took your second option: renamed to **"Toggle Looping"** and driven from one `self.loop_enabled` flag (initialised `True`, and `video_container` now seeds each pane from it). This keeps all four widgets the same type and preserves the plan's prescribed uniform `list[tuple[str, Callable]]` shape, while fixing both defects — the label no longer describes the opposite of a first click, and mixed states converge instead of inverting per-pane forever. Test: `test_toggle_looping_converges_on_mixed_state` desynchronises two panes and asserts they agree after each toggle (fails pre-fix with the per-pane version).

### 5. B4 regression test added
`test/test_utils.py::TestPublishFileContract` — `inspect.signature(publish_file).return_annotation` is `"None"` (string, per `from __future__ import annotations`), plus an AST walk asserting no `return <expr>` in the body. No git remote touched. **Verified failing** against the pre-fix `-> str` by checking out `398bdd96:bencher/utils.py` in place: `AssertionError: publish_file claims -> str`. Plan §9 DoD now literally satisfied for B1–B5.

### 6. CHANGELOG + `schema_version` policy documented
`CHANGELOG.md` `[Unreleased]` → `### Fixed` now carries B5, B1, B4 and the crash path, and the B5 entry states the recolouring outright: no version gate in `discover.py:56-64`, so every existing `*.summary.json` is re-read with the corrected rules and `adaptive`/`delta`/`absolute` cells can move between `improved` and `passed` with no file changing and no benchmark re-running (`regressed`/`trend` unaffected). `docs/scorecard.md` gains a note declaring `schema_version` a **structural** version — shape, not value semantics — which is the policy gap behind the bump question, documented rather than bumped.

### 7. PR body evidence claim amended
Rewritten to say what the 13 failures actually are: 7 vacuous (3 × `TypeError: _verdict() missing 3 required positional arguments`, 4 × `AttributeError`), with the behavioural evidence being the 2 button tests plus the 4 `TestCellVerdictMethodUnits` cases that go through the unchanged `cell_verdict` signature and show real verdict diffs.

### 8. Cleanliness
- `_finite_value(value: object)`, and `Mapping[str, object]` on `_verdict` / the three helpers (this file is on the plan's ty-hardening path).
- Kept `_finite_value` separate from `_finite_or_none` rather than collapsing them, and documented why: tolerant **reader** for summaries bencher may not have written vs strict **writer** for values handed to `to_dict`, where a non-numeric input is a construction-time programming error. Collapsing them would have turned that into a silent `null`.
- Delta boundary made strict (`>`) to mirror `detect_delta`'s `delta > max_delta` exactly, since the adaptive branch is exact; documented that `_percent_improved` keeps its pre-existing inclusive `>=` (all four detectors use strict `>`, so that inclusive convention predates this PR and changing it is out of B5's unit-mismatch scope).

### Verification
`TMPDIR=$(mktemp -d) PYTEST_DEBUG_TEMPROOT=$TMPDIR pixi run ci`: format clean, ruff clean, **pylint 10.00/10**, **`ty` all checks passed**, **2615 passed / 3 skipped**, coverage 90%. The 4 new behavioural tests were each confirmed failing against the previous commit.

PR #1037, opened 2026-07-31
refactor: JobFuture Ready|Pending sum type and always-hashed WorkerJob (plan 23 P5)

blooop:
Implements plan 23 P5 (C2 + C8) — `plans/23-constructive-data-modeling.md:207` (C2), `:247` (C8), `:513` (P5). Implementation notes recorded in that plan's §10 as `### P5 (implemented)`.

## C2 — `JobFuture` holds one state, not two optionals

`JobFuture.__init__` stored `res` and `future` as independent optionals and `result()` *mutated* `res` while leaving `future` set, so `future is not None` stopped meaning "pending" after the first call and both-set / neither-set were both representable.

- `bencher/job.py:233-268` — `Ready(res: dict)` | `Pending(future: Future)` | `Broken(error: WorkerReturnedNothingError)`, frozen dataclasses, aliased as `JobState`.
- `bencher/job.py:292` `JobFuture.__init__` — the keyword signature is preserved (`job`, `res`, `future`, `cache`), so `FutureCache.submit`'s sites and the three hand-built test objects read as before; the arguments are parsed into one `state` field at the boundary. At most one arm may be given — previously `res` and `future` together were representable and never meaningful.
- `bencher/job.py:341` `JobFuture.result()` — exhaustive `match` + `assert_never`, and **total**: the P2-era `-> dict | None` is gone.
- `bencher/bencher.py:1205` `Bench.calculate_benchmark_results` — the order-dependent `job_future.future is not None` read now matches on `job_future.state`.

### How this reconciles with #1029/#1032 (the part of P5's spec that had to change)

P5 says *"the worker-`None` check from P2 moves into the factory that constructs `Ready`"*. **Taken literally that reinstates the original B3 bug** and I did not do it: the serial construction site (`FutureCache.submit`) runs *inside* the caller's `except catch` block, so a `raise` at construction is absorbed by `catch=Exception` — precisely the failure mode plan 23 P2 item 1 measured and `TestCatchDoesNotChangeContractHandling` pins.

So the check lives at construction as a **discovery**, never as a disposition: a job with no result parses to `Broken`, which *stores* the error and raises it from `result()` at the consume point. **The #1032 disposition is unchanged** — `store_results` records a `SampleFailure`, logs at ERROR, emits `WorkerContractWarning`, leaves the cells at the missing sentinel, and the sweep continues. A `None` return still cannot abort a run on either executor path.

Two consequences worth flagging for review:

1. **`Broken` is a third arm the plan text does not name.** With only `Ready | Pending` the None case has to be either `Ready(None)` (the sentinel back again) or a raise at construction (above). `Broken` isn't invented state — it is the one *meaningful* half of the "neither set" case the old shape allowed, given a name. Its error is *supplied* by whichever caller knows the cause (`FutureCache.submit`'s serial site at `bencher/job.py:625`, having just watched `run_job` return `None`) rather than inferred from the shape of the constructor call, which cannot tell a `None`-returning worker from a cached `None`. `Broken.error` is typed `WorkerReturnedNothingError`, so a `Broken` carrying a worker's own exception is unrepresentable too.

2. **One ordering change in `store_results` (`bencher/result_collector.py:443-476`) is load-bearing, and the handler must be a NARROW subclass.** `require_worker_result` used to run *after* the `try/except catch`; now `result()` raises the violation from *inside* that `try`. The error subclasses `TypeError`, so `except catch` with `catch=Exception` would absorb it — the handler is therefore ordered **first**. Verified by swapping the two: `TestCatchDoesNotChangeContractHandling::test_catch_does_not_swallow_a_none_return` fails.

   Review caught the other half of this: catching the *public base class* `WorkerContractError` there silently tolerated a **worker-raised** one on the pooled path even with `catch=()` — aborting on SERIAL but completing with `n_failed=2` on MULTIPROCESSING, i.e. loud-or-silent chosen by the executor, the exact defect shape B3 exists to kill. Fixed by splitting the vocabulary: `WorkerReturnedNothingError` (`bencher/job.py:139`) is the harness's own diagnosis and the only thing exempt from `catch=`; a worker-raised `WorkerContractError` falls through to `except catch` as it did before P5. `TestAWorkerRaisedContractErrorIsStillASampleFault` (`test/test_collection_contract.py:279`) pins all four cells and was verified to fail on precisely the `pool` + `catch=()` cell when widened back to the base class. **The lesson: the disposition split is a property of the exception vocabulary, not of handler placement.**

`assert_never` is licensed by plan 24 A1 — the subject is `JobFuture.state`, a field this module writes and annotates, not a `param` read. Measured, not assumed: deleting the `Broken` arm gives `error[type-assertion-failure] … Inferred type of argument is 'Broken & ~Ready & ~Pending'`. Conversely `bencher.py`'s dispatch deliberately **does not** end in `assert_never` — `job_future` arrives from an untyped `list[tuple]` (A1: exhaustiveness there would be an assertion, not a proof), and its default arm is real behaviour rather than a residual.

Caching semantics preserved exactly: the pre-P5 `if self.cache is not None and self.res is not None` guard becomes `_cache_and_return`, reachable only with a non-`None` dict — enforced by construction rather than by a guard.

## C8 — `WorkerJob` cannot exist with unset hashes

`bencher/worker_job.py` — `function_input`, `canonical_input`, `fn_inputs_sorted` and `function_input_signature_pure` defaulted to `None` and were filled by `setup_hashes()`, which every construction site had to remember to call; nothing prevented caching a sample under `job_key=None`. They are now `cached_property`s and `setup_hashes()` is gone (its one production caller was in `bencher.py`'s job-building loop).

`cached_property` rather than the factory classmethod the plan also offers: a factory leaves the plain constructor reachable, which *was* the defect, and `__post_init__` would recompute hashes on every unpickle.

**Cache safety: no `CACHE_VERSION` bump.** `function_input_signature_pure` is still `hash_sha1((sorted(function_input.items()), tag))` over the same `function_input`, and `canonical_input` is still computed from the swept dims *before* constants are merged — the ordering the old imperative method relied on, now two independent properties that each rebuild the pre-merge dict. Proven byte-identical against `origin/main` for all four derived values over 10 edge cases (including `None` values, a list value, a const shadowing a swept dim, and an all-const job), and independently re-proven by review with an end-to-end sweep written by `main`'s code and read by this branch's → 6/6 cache hits.

`found_in_cache` and `msgs` are **deleted**: nothing in `bencher/` reads or writes either, and the only references were `test_worker_job.py`'s own default-checking test.

## Strict `ty` list: succeeded, and for three files rather than two

`pyproject.toml:320-331` — `bencher/job.py`, `bencher/worker_job.py` **and** `bencher/result_collector.py`. `pixi run ty` is `All checks passed!` with all three listed, so the block's "added only once clean" rule is honoured, not loosened.

`result_collector.py` is the bonus: plan 23 P2 item 7 recorded 4 remaining diagnostics and predicted *"fixing [WorkerJob's two-phase init] in P5 should let both files onto the strict list in one step"*. **Confirmed** — both `function_input`-is-`None` diagnostics were the same defect C8 removed. Getting there also surfaced two live defects P2 had recorded as out of scope, and in both cases the fix goes **past** what the checker asked for, because the narrowest change would have hidden the defect it pointed at:

- `bencher/job.py:655` `FutureCache.clear_tag` called `self.cache.evict(tag)` unguarded on `Cache | None` — an `AttributeError` for anyone with `cache_samples=False` clearing a tag. Now guarded *and* logging a WARNING: on a public path (`Bench.clear_tag_from_sample_cache`) a silent no-op must not read as "the tag was cleared".
- `bencher/job.py:743` `JobFunctionCache.call` passed `self.call_count` (an `int`) as `Job.job_id`, declared `str`. But `call_count` was initialised to 0 and incremented **nowhere**, so every call produced the same job id — the string every log line and contract message identifies a sample by. Casting alone would have made the annotation true while preserving that; it is now incremented (`call 1`, `call 2`, …). Cache keys are unaffected: `Job.job_key` derives from `job_args`, never `job_id`.

## Falsified plan claims

1. **P5's "the worker-`None` check moves into the factory that constructs `Ready`" is wrong as written** — see above. The check's *location* moved as the plan intended; its *disposition* could not, and a literal reading reintroduces the bug #1029/#1032 fixed.
2. **P5's `Ready | Pending` is not sufficient** — the union needs a third arm for `result()` to be total without a sentinel.
3. **P5's DoD understates the strict-list win** — it names two files; three are clean, exactly as P2 item 7 predicted.
4. **The pre-P5 `dict | None` docstring on `result()` claimed only P5's split could remove the `| None`.** True, and now done — but the removal came with the `Broken` arm the docstring did not anticipate.
5. **P5's own first implementation claimed `WorkerJob` crosses a process boundary; it does not.** Only `Job` (function + `job_args`) is submitted to an executor — a `WorkerJob` is built and consumed in the parent. The `cached_property` choice stands on its other reasons; the claim is corrected in the docstring, the test comment, and plan §10.
6. **This plan cited a test class that does not exist** — `TestCatchDoesNotAbsorbContractErrors` at two places (one pre-existing, one added here). The class is `TestCatchDoesNotChangeContractHandling`. Per plans-README rule 7 the symbol is the durable reference, so a plan whose thesis is "claims are measured" should not cite one that isn't there.

## Tests

`test/test_sample_fault_tolerance.py`, `test/test_sample_cache.py` and `test/test_cache_management.py` are **unchanged** — they pin the #1032 disposition and were the acceptance criterion for `Broken`'s semantics.

`test/test_collection_contract.py` gains the review's matrix (`TestAWorkerRaisedContractErrorIsStillASampleFault` + a module-level `RaisesContractError` worker so it pickles for the pool). **One existing assertion in it changed**, flagged rather than buried: `test_catch_does_not_swallow_a_none_return` did a `repr()` substring check for `"WorkerContractError"`, and the recorded class is now the narrower `WorkerReturnedNothingError`. It asserts the narrower name — strictly stronger, and it keeps the property the test exists for. Its wrong-length-vec sibling still checks the base class and still passes, since `_store_result_vars` raises that.

Otherwise changed only where the removed API was referenced:
- `test/test_worker_job.py` — `make_job` drops its `setup_hashes()` call; `test_defaults_before_setup` (asserting the four fields are `None`) is replaced by `test_hashes_exist_from_construction`; `test_msgs_lists_are_independent` (a `default_factory` regression test for a field that no longer exists) is dropped with `msgs`.
- `test/test_multiprocessing_executor.py` — `setup_hashes()` call replaced by touching `function_input_signature_pure`, so the pickle round-trip covers *populated* cached values rather than only lazy recomputation. Kept as a property test, with its "sent alongside the Job" comment corrected.

New in `test/test_job.py`: `TestJobFutureState` (variant parsing; every conflicting pair rejected; `res={}` still counts as given; the variant does not change on resolve — the property `bencher.py`'s old read lacked; caching on both paths), `TestJobFutureNoneReturn` (`Broken` stays quiet at construction and raises at the consume point; the generic diagnosis does not blame the benchmark function; an explicit error is kept verbatim; never cached; a worker-raised `WorkerContractError` is *not* the harness diagnosis), `TestJobFunctionCacheJobIds` and `TestClearTagWithoutACache`.

Verified with a private temp root (`TMPDIR=$(mktemp -d) PYTEST_DEBUG_TEMPROOT=$TMPDIR`): `pixi run ci` **2616 passed, 3 skipped**, no collection errors; `pixi run test-split` 2616 passed; pylint 10.00/10; `ty` clean; `prek` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,10 +68,54 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   back. The new `Protocol` (`submit`, `shutdown`) is satisfied by both
   `ProcessPoolExecutor` and scoop's module.
 
-- `JobFuture.result()` is annotated `-> dict | None`. The `None` is not new behaviour, only
-  newly admitted: `JobFuture` can represent "no result and no future", and callers must
-  reject it. Plan 23 P5's `Ready(dict) | Pending(Future)` split is what makes the state
-  unrepresentable and the `| None` removable.
+- `JobFuture.result()` was annotated `-> dict | None` by P2 (the `None` was not new
+  behaviour, only newly admitted: `JobFuture` could represent "no result and no future").
+  **Superseded before release by P5 below**, which makes that state unrepresentable and
+  the method total.
+
+- **`JobFuture` holds one state field instead of two optionals** (plan 23 P5, C2). `res`
+  and `future` were independent optionals that `result()` *mutated* — it assigned to
+  `self.res` and left `self.future` set — so `future is not None` stopped meaning
+  "pending" after the first call, and both-set and neither-set were both representable
+  while neither was meaningful. The field is now a single
+  `Ready(res) | Pending(future) | Broken(error)` sum of frozen dataclasses, parsed once in
+  the constructor (whose keyword signature is unchanged, so every construction site and
+  hand-built test object reads as before); passing both a result and a future now raises
+  `ValueError`. `result()` matches exhaustively with `assert_never` — verified by deleting
+  an arm and measuring `error[type-assertion-failure] … Inferred type of argument is
+  'Broken & ~Ready & ~Pending'` — and is **total**: the `| None` return is gone. The
+  order-dependent `job_future.future is not None` dispatch in `Bench.calculate_benchmark_results`
+  reads the variant instead.
+
+  `Broken` is the constructive replacement for the one meaningful half of "neither set": a
+  serial worker that returned `None`. The error is *stored* rather than raised at
+  construction, because the serial site is inside the caller's `except catch` block —
+  raising there is exactly the original B3 failure mode where `catch=Exception` absorbed a
+  contract violation. **The record-and-continue disposition is unchanged** (plan 23 §6.2
+  as amended): `result()` raises `WorkerContractError` at the consume point on *either*
+  executor path, `store_results` records a `SampleFailure`, logs at ERROR, emits
+  `WorkerContractWarning`, and the sweep continues. A `None` return never aborts a run.
+  Caching semantics are also unchanged: the pre-P5 `res is not None` guard on `cache.set`
+  is now enforced by construction — a worker that returned nothing cannot reach the cache.
+
+- **`WorkerJob`'s derived inputs and hashes are `cached_property`s, not two-phase init**
+  (plan 23 P5, C8). `function_input`, `canonical_input`, `fn_inputs_sorted` and
+  `function_input_signature_pure` defaulted to `None` and were filled by a `setup_hashes()`
+  call every construction site had to remember, so nothing prevented caching a sample under
+  `job_key=None`. `setup_hashes()` is gone; the four values are computed on demand from the
+  constructor fields, so a `WorkerJob` with unset hashes no longer exists. The unused
+  `found_in_cache` and `msgs` fields (written nowhere in the package) are removed. Hash
+  values are byte-identical — the same `hash_sha1((sorted_inputs, tag))` over the same
+  inputs — so no cached sample is invalidated, and pickling across a process boundary still
+  works (computed values travel in `__dict__`; unaccessed ones recompute deterministically).
+
+- `FutureCache.clear_tag` no longer raises `AttributeError` on `None` when the cache is
+  disabled (`cache_samples=False`), and `JobFunctionCache.call` passes its counter as a
+  `str`, matching `Job.job_id`'s declared type. Both were latent defects surfaced by putting
+  `job.py` on the strict `ty` list (recorded as out-of-scope in plan 23 P2 item 7).
+  `bencher/job.py`, `bencher/worker_job.py` and `bencher/result_collector.py` are all on
+  that list now — the last one because C8 removed the `function_input`-is-`None` diagnostics
+  that were holding it off, exactly as P2 predicted.
 
 ### Removed
 - **BREAKING: dropped Python 3.10 support.** `requires-python` is now `>=3.11,<3.14`, and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,12 +88,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   reads the variant instead.
 
   `Broken` is the constructive replacement for the one meaningful half of "neither set": a
-  serial worker that returned `None`. The error is *stored* rather than raised at
-  construction, because the serial site is inside the caller's `except catch` block —
-  raising there is exactly the original B3 failure mode where `catch=Exception` absorbed a
-  contract violation. **The record-and-continue disposition is unchanged** (plan 23 §6.2
-  as amended): `result()` raises `WorkerContractError` at the consume point on *either*
-  executor path, `store_results` records a `SampleFailure`, logs at ERROR, emits
+  job that yielded no result. The error is *stored* rather than raised at construction,
+  because the serial site is inside the caller's `except catch` block — raising there is
+  exactly the original B3 failure mode where `catch=Exception` absorbed a contract
+  violation. It is supplied by whichever caller *knows* the cause (the serial site in
+  `FutureCache.submit`, having just watched `run_job` return `None`) rather than inferred
+  from the shape of the constructor call, which cannot tell a `None`-returning worker from
+  a cache entry holding `None`. **The record-and-continue disposition is unchanged** (plan
+  23 §6.2 as amended): `result()` raises at the consume point on *either* executor path,
+  `store_results` records a `SampleFailure`, logs at ERROR, emits
   `WorkerContractWarning`, and the sweep continues. A `None` return never aborts a run.
   Caching semantics are also unchanged: the pre-P5 `res is not None` guard on `cache.set`
   is now enforced by construction — a worker that returned nothing cannot reach the cache.
@@ -117,6 +120,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   that list now — the last one because C8 removed the `function_input`-is-`None` diagnostics
   that were holding it off, exactly as P2 predicted.
 
+  Both fixes go past what the type checker asked for, because in both cases satisfying it
+  would have hidden the defect it pointed at. `FutureCache.call_count` was initialised to 0
+  and incremented **nowhere**, so every `JobFunctionCache.call()` produced the same job id —
+  the string every log line and contract message identifies a sample by; it is now
+  incremented, and ids read `call 1`, `call 2`, …. And `clear_tag` on a cache-less
+  `FutureCache` now logs a WARNING rather than returning silently: not crashing is right, but
+  on a public path (`Bench.clear_tag_from_sample_cache`) "nothing happened" must not read as
+  "the tag was cleared".
+
 ### Removed
 - **BREAKING: dropped Python 3.10 support.** `requires-python` is now `>=3.11,<3.14`, and
   the CI matrix runs py311 + py313 (was py310 + py313). The `py310` pixi feature and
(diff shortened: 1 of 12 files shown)


blooop:
## Review addressed — all six findings, in `66999c1e`

Thank you for the 12-case disposition matrix and the two independent cache-key proofs; the MEDIUM finding is real and was a genuine behaviour regression I had not measured. Details per item.

### 1. MEDIUM — worker-raised `WorkerContractError` silently tolerated on the pooled path — **fixed**

Reproduced your matrix row exactly. Took your suggested fix.

- `bencher/job.py:139` — new `class WorkerReturnedNothingError(WorkerContractError)`: the harness's *own* diagnosis, never a worker's.
- `bencher/job.py:156` `_returned_none_error` and `:170` `_no_result_error` both mint the narrow subclass.
- `bencher/result_collector.py:465` — the handler ahead of `except catch` is now `except WorkerReturnedNothingError`.

`TestAWorkerRaisedContractErrorIsStillASampleFault` (`test/test_collection_contract.py:279`) pins all four cells (`{SERIAL, MULTIPROCESSING}` × `{catch=(), catch=Exception}`) plus the subclass relation, driven by a new module-level `RaisesContractError` worker so it pickles for the pool.

**Verified the test detects the regression, not just the fix.** Widening the handler back to `except WorkerContractError` and re-running:

```
FAILED …::TestAWorkerRaisedContractErrorIsStillASampleFault::test_no_catch_aborts_on_either_executor[pool]
1 failed, 4 passed
```

Exactly the one cell your table identified — serial still aborts, pool no longer does.

Every existing `assertRaises(WorkerContractError)` still passes by subclassing, and `_store_result_vars`' three raises are outside this `try` and untouched, as you noted. **One assertion in `test_collection_contract.py` did have to change**, and I want to flag it rather than bury it: `test_catch_does_not_swallow_a_none_return` asserted `"WorkerContractError" in failed_samples[0].exception`, which is a `repr()` substring check — the recorded class is now `WorkerReturnedNothingError`, which does not contain that substring. I made it assert the narrower name, which is strictly *stronger* (it pins that the recorded failure is the harness diagnosis specifically) and keeps the property the test exists for. The wrong-length-vec sibling assertion still checks `WorkerContractError` and still passes, since `_store_result_vars` raises the base class.

The generalisable lesson, now recorded in the plan (§10 item 11): **the disposition split is a property of the exception vocabulary, not of handler placement.** P5 moved a handler and silently changed which errors it governed, because one class was doing two jobs.

### 2. LOW — `str()` cast satisfied ty but preserved the defect — **fixed**

Confirmed: `call_count` was set to 0 at `__init__` and incremented nowhere, and nothing else in `bencher/` or `test/` reads it. Now incremented in `JobFunctionCache.call` (`bencher/job.py:743`) with ids reading `call 1`, `call 2`, … Measured before/after:

```
before: job_ids: ['0', '0', '0']
after:  job_ids: ['call 1', 'call 2', 'call 3']
```

Cache keys are unaffected — `Job.job_key` derives from `job_args`, never `job_id`. Pinned by `TestJobFunctionCacheJobIds`. Also corrected the `call_count` docstring, which described it as a general counter.

You're right that this is a failure mode the strict ratchet actively invites: the checker asks for the narrowest possible change, and the narrowest change is often the one that hides what it found. Recorded as plan §10 item 12.

### 3. LOW — `clear_tag` guard was a silent no-op with no test — **fixed**

Now logs a WARNING naming the tag (`bencher/job.py:655`), and pinned by `TestClearTagWithoutACache` — one test that it does not raise (the pre-existing `AttributeError`), one that it says so.

### 4. LOW — `Broken` inferred a cause from a syntactic condition — **fixed**

You were right that "neither `res` nor `future`" cannot distinguish a `None`-returning worker from a cached `None`. Took the first of your two options:

- `JobFuture.__init__` gains an optional `error=` (`bencher/job.py:292`), so the caller that *knows* the cause supplies it.
- `FutureCache.submit`'s serial site (`bencher/job.py:625`) now does the `None` check explicitly and passes `_returned_none_error(...)` — it is the only place that has watched `run_job` return `None`, so it is the only place that says so.
- The no-argument fallback gets `_no_result_error`, which claims only what is known and no longer blames the benchmark function.
- The arm-conflict check now spans all three arms and uses `is not None` rather than truthiness — `res={}` is a valid result and must still count as given. `TestJobFutureState` parametrizes all three conflicting pairs plus that empty-dict case.

`Broken.error` is narrowed to `WorkerReturnedNothingError`, so "a `Broken` carrying a worker's own exception" is now unrepresentable too.

### 5. LOW — inaccurate pickling justification — **fixed (claim, not code)**

Confirmed your reading: only `Job` (function + `job_args`) is submitted to an executor; the `WorkerJob` is built and consumed in the parent. Corrected in all three places — the `WorkerJob` docstring, `test/test_multiprocessing_executor.py`'s section comment and class docstring (test kept; it is a good property test that every value type a sweep produces survives a round trip), and plan §10 item 6 via a new item 14. The `cached_property`-over-`__post_init__` choice stands on its other two reasons.

### 6. NIT — nonexistent test class cited — **fixed**

`TestCatchDoesNotAbsorbContractErrors` → `TestCatchDoesNotChangeContractHandling` at both `plans/23-constructive-data-modeling.md:852` (pre-existing) and `:933` (mine), and in the PR body. Fair hit on a plan whose thesis is that claims are measured.

### Re-verification

Per your instruction, with a private temp root (`TMPDIR=$(mktemp -d) PYTEST_DEBUG_TEMPROOT=$TMPDIR`):

- `pixi run ci` — **2616 passed, 3 skipped**, no collection errors (2602 → 2616 is the 14 new tests). Pylint 10.00/10.
- `pixi run test-split` — 2616 passed, 3 skipped.
- `pixi run ty` — `All checks passed!` with all three files on the strict block.
- `pixi run prek` — clean.
- **Cache keys re-proven byte-identical to `origin/main`** after these changes: all four derived values over 10 edge cases (including `None` values, a list value, a const shadowing a swept dim, and an all-const job) diff-clean against a detached `origin/main` checkout.

Also worth noting: the private temp root explains the 88 transient collection errors I reported earlier as unreproducible — they have not recurred under it.

Not actioned, per your note: cached-property staleness through pickle (identical hazard to `setup_hashes`, and nothing mutates a `WorkerJob` after construction), and `hash(Ready({}))` raising `TypeError` (unreachable — the pooled dispatch keys by the `Future`).

Not merging — over to you.

blooop:
Note for the record: this merged at `66999c1e`, which contains all six review fixes. One cosmetic commit (`72fd90dc`) reached the branch a few minutes after the merge and was left out — a stale docstring line count and `Broken(error or ...)` → an explicit `elif error is not None` arm. Carried over in #1042; no behaviour change, safe to close if not wanted.

PR #1030, opened 2026-07-31
feat: single ResultSpec registry replaces the nine result-type tuples (plan 23 P4)

blooop:
Implements **plan 23 P4** (`plans/23-constructive-data-modeling.md` §D3/§P4): one ordered
`RESULT_SPECS: dict[type, ResultSpec]` mapping in `bencher/variables/results.py` now declares
every `Result*` class's kind, missing fill/sentinels, and family memberships. The nine
hand-maintained registries (C1) are **derived under their existing names** — zero call-site
churn: `PANEL_TYPES`, `SCALAR_RESULT_TYPES`, `XARRAY_MULTIDIM_RESULT_TYPES`,
`ALL_RESULT_TYPES`, `RESULT_KIND_ORDER`, `_REFERENCE_MISSING_TYPES`, `_OBJECT_MISSING_TYPES`,
`DATA_VAR_RESULT_TYPES`, and `result_collector`'s `_MEDIA_RESULT_TYPES` (now imported, not
redeclared; PR #1028's aging-rationale comment moved with it).

This is the prerequisite A6 phase 2 (grammar of ND data) consumes: a closed channel
vocabulary derives from `RESULT_SPECS` rather than nine tuples (plan 23 §8).

## DoD

- **Adding a `Result*` class without a spec fails CI with one clear message** — verified by
  mutation: a probe `ResultProbe` class fails
  `TestResultSpecRegistry::test_every_result_class_is_registered_or_exempt` naming the class
  and the fix.
- The misclassification hole is closed: a parameter class defined in
  `bencher.variables.results` but absent from the registry now **raises at
  sweep-declaration time** (`parametrised_sweep.py`) instead of silently becoming an
  *input* variable (the old `ResultVolume` trap).
- `bencher/variables/results.py` added to the strict ty ratchet after fixing its one Tier-C
  diagnostic (`ResultReference.__init__`'s malformed `Callable[Any, pn.pane.panel]`
  annotation — one of the handover §4 shapes). `pixi run ty`: All checks passed.
- Post-plan-22 fills preserved exactly: `ResultDataSet` keeps its dual-generation sentinels
  (`"NAN"` + legacy `-1`) as a **documented per-spec `missing_sentinels` set plus the
  unchanged `_dataset_cell_is_missing` read path** — not collapsed, per D3.
- `ResultHmap` reproduces its historical quirks (registered; neither data-var nor panel);
[...]

PR #1038, opened 2026-07-31
docs: plan 25 — A6 phase 2, channel plan derived from the result-type registry

blooop:
Docs only. Writes `plans/25-grammar-phase-2-channel-plan.md` — the A6 phase-2
implementation plan — and adds its index row to `plans/README.md`.

Written against `main` @ `398bdd96`, i.e. **after** plan 23 P4, exactly as plan 23 §8
directed: "the phase-2 plan should be written against the post-P4 registry — a closed
channel vocabulary is much easier to derive from `RESULT_SPECS` than from nine tuples."

## What the plan proposes

Four independently-shippable PRs, additive and **shadow-only** (zero visual change; the
acceptance artifact is a byte-identical gallery):

| Phase | Content |
|---|---|
| **P1** | `Channel` (9 members, closed per Law 5), `Frame`, `SpreadStat`, and `kind_caps()` — a total function over `ResultKind` with `assert_never`, so a new `Result*` class cannot reach the grammar without capabilities |
| **P2** | Frozen, picklable, canonical-JSON `Plan` / `Reject` / transform records, `PlanInputs` boundary parse, `MarkDecl` seed table keyed on existing plugin names |
| **P3** | The planner: A6 Law 7 policy v1.1 step-by-step, cardinality budgets with demote-then-reject, `explain()`, determinism pinned three ways |
| **P4** | Shadow harness against the real pathways, a divergence ledger where every disagreement cites a numbered A6 §6 finding, and the repo's first golden-file corpus |

The central design move is the one plan 23 §8 asked for: channel legality is **derived**
from `RESULT_SPECS`/`ResultKind` rather than redeclared, so the chain "new result class →
spec → kind → capabilities" has no silent link.

## Owner principles honored

- **Never crash mid-run.** D6 makes the planner a total function returning
  `Plan | Reject`, and states the phase-3 integration contract once: any `Reject` or
  exception on the auto path degrades through the existing
  `report_render_failure`/`RenderFailedWarning` pattern (`bencher/results/render_failure.py:34-58`),
  mirroring the worker-contract disposition from plan 23 §10's P2 amendment. `Reject`
  carries `reasons` + `suggestions` so it *is* the visible pane without reformatting,
  and reject cases are in the golden corpus from birth.
- **Correctness over convenience.** No `override=` anywhere in the grammar package
  (grep-enforced), the planner never applies a data transform the user did not ask for,
  and divergences from today's behavior are ledgered and reviewed rather than absorbed.

## A6 claims found stale (recorded in the plan's §2.6, not silently worked around)

1. A6's header says "No implementation yet" — phase 1 has landed (plans 22 + PRs #1019/#1021/#1022/#1030-#1032).
2. `hmap_kdims = sorted(...)` cited at `bencher.py:1059` → now `bencher/bencher.py:1107`.
3. The "rerun over-time grid" pathway is placed among the rerun/video files → it is `_pane_over_time_grid`, `bencher/results/bench_result_base.py:1066-1103`.
4. Law 1 says `ResultHmap` "is deprecated and removed" → deprecated only; removal is phase 3, so phase 2 must plan around live hmaps.
5. "Nine pathways / three dialects" are floors, not exact counts — the surface has grown three ways (a tenth over-time-grid member `_pane_over_time_dataset`, created by phase 1 itself; the `TabularSpec`/xy family; `RerunSummaryResult`), and there is now a fourth gate layer (the plugin registry's `requires`/priority ladder).
6. Finding #21's "three divergent nesting orders" has a fourth instance: `rerun_summary.py` reproduces the video `reverse`-not-propagated bug (`:202`, `:217-218`, `:238-244`), unlisted in the audit.
7. Law 5's entry-point group name is correct (`plugins/registry.py:14`), but `pyproject.toml` declares no entry points — builtins register imperatively, so the group is a third-party surface only.

All nine original pathways and all three filter dialects were re-verified present, and
every A6 §6 finding the plan's harness leans on was spot-checked live (§2.4).

## Deviations from A6, recorded with evidence (§9)

`ResultVec` index-as-dim is deferred to phase 3 (it is a stored-data change misfiled
under planner invariants, and phase 2 may not touch stored data); backend capability
tables are scoped to panel/holoviews until the lowerings can validate them; plan-at-collect
storage waits for its first consumer. Each has an OWNER DECISION entry with the
alternative recorded.

## Validation

`pixi run ci` green (2591 passed, 3 skipped) and `pixi run prek` green including
codespell. Note: an initial `ci` run showed 88 setup errors from a shared-`/tmp`
collision (`FileNotFoundError: /tmp/pytest-of-.../pytest-1`, a concurrent pytest session
removing the basetemp) — unrelated to this change and clean under an isolated `TMPDIR`.

## Summary by Sourcery

Document the A6 grammar phase 2 implementation plan, introducing the channel vocabulary, plan/planner design, and shadow-planning strategy, and index it in the plans README.

Documentation:
- Add plan 25 document describing grammar phase 2: closed channel vocabulary, Plan/Reject types, planner policy, shadow harness, and golden corpus strategy for A6.
- Update plans index to include plan 25 with its scope, risk/size, and relationship to other A6 stack phases.

### DIFF
--- a/plans/25-grammar-phase-2-channel-plan.md
+++ b/plans/25-grammar-phase-2-channel-plan.md
@@ -1,0 +1,829 @@
+# Plan 25 — Grammar Phase 2: Channel Vocabulary, `Plan` Type, and Shadow Planner
+
+**Goal:** Implement phase 2 of the A6 migration
+([A6 — Grammar of ND data](architecture/A6-grammar-of-nd-data.md), Law 10 step 2):
+the closed dimension→channel vocabulary (Law 5) derived from the `RESULT_SPECS`
+registry, the frozen `Plan` type and transform records (Laws 2/5/7), a deterministic
+planner implementing default assignment policy v1.1 (Laws 6/7) with `explain()`, and a
+**shadow-mode** harness that asserts agreement with today's selection across the whole
+gallery — every divergence cited to a numbered A6 §6 finding. The golden-plan corpus is
+born here, before any renderer changes. **Zero rendering changes; zero visual changes**
+— phase 3 is the first phase allowed to alter a pixel.
+
+**Branch names:** one `plan/grammar-p2-*` branch per phase (each phase is one PR).
+
+**Citations pinned to:** `main` @ `398bdd96` (post plan 23 P4 / PR #1030 and the P2
+warn-not-crash amendment / PR #1032, 2026-07-31). Per plans-README rule 7, confirm each
+`file:line` against the current tree before relying on it; the symbol is the durable
+reference.
+
+**⚠️ Read first:** A6 §2 Laws 5–7 and Law 10 (this plan implements exactly its phase-2
+step), A6 §5–6 (the empirical baseline and the 24 findings — the shadow harness's
+oracle), plan 22 §"Amendments" (what phase 1 actually became), and plan 23 D3/P4 (the
+`RESULT_SPECS` registry this plan derives from). §2.6 below lists where A6's own
+citations have gone stale; this plan's §2 is the refreshed baseline.
+
+**Two owner principles bind every design choice here** (both resolved on the record):
+
+- **(a) bencher never crashes mid-run** — plan 23 §10 "P2-amendment" (owner decision,
+  2026-07-31): failures degrade visibly, never abort. The rendering-side mechanism
+  already exists: `RenderFailedWarning` + `report_render_failure`
+  (`bencher/results/render_failure.py:30-58`, PR #1027) — an ERROR log, a warning a
+  strict pipeline can promote, and a labelled failure pane in the report. The planner
+  is designed so this principle is structural, not aspirational (D6).
+- **(b) long-term correctness beats short-term convenience** — no shims that would let
+  a wrong plan render "close enough"; divergences from today's behavior are recorded
+  and reviewed, not absorbed.
+
+---
+
+## 1. Scope and non-goals
+
+In scope (all additive; no existing module's behavior changes):
+
+- **`bencher/grammar/`** — a new package: channel vocabulary, kind→channel capability
+  derivation from `RESULT_SPECS`, frozen `Plan`/transform/mark-declaration record
+  types, the planner, `explain()`. On the strict `ty` list from day one (plan 23 D1 —
+  new code has no legacy excuse).
+- **Shadow harness + divergence ledger** — test-only (A6 Law 10: "Shadow mode exists
+  only inside phase 2's test suite").
+- **Golden-plan corpus** — checked-in serialized plans with a regeneration task and a
+  CI diff check. This repo has no snapshot infrastructure (§2.7); this plan creates the
+  first, following the documented-update-procedure precedent of
+  `TestGoldenBenchCfgHash` (`test/test_hash_persistent.py:766-807`).
+
+Out of scope (recorded so nobody "helpfully" pulls them forward):
+
+- **Any renderer change.** `_to_panes_da`, `map_plot_panes`, the composable containers,
+  rerun, video — untouched. Phase 3 rewrites panel/holoviews as plan execution; phase 4
+  does rerun/video. The DoD greps enforce that nothing under `bencher/results/` or
+  `bencher/plugins/` imports `bencher.grammar` (§8).
+- **The Law 8 API surface** (`res.view()`, `views=`, `bn.compose`) — phase 5.
+- **Plan-at-collect storage** (Law 9's stored plans). Deferred to phase 3 with
+  rationale: a stored plan with no consumer is dead weight plus a compatibility surface
+  for old pickles; nothing in Law 9 requires storing before a renderer reads plans.
+  Law 9's *cache* half — nothing plan-shaped ever enters a cache key — is enforced
+  starting now (§7).
+- **Backend capability tables for plotly/video/rerun** (Law 3). Phase 2 ships the
+  capability-table *type* and one seed table for the panel/holoviews backend — the
+  minimum the planner needs to shadow today's auto path. The other backends' tables
+  land with their lowering phases (3–4), where they can be validated against real
+  lowerings instead of guessed.
+- **Opening the mark registry to entry points** (Law 5's "marks are open"). Phase 2's
+  mark declarations are an internal table keyed by the existing plugin names
+  (`bencher/plugins/builtins.py:66-139`) so phase 3 can migrate name-by-name; the
+  public entry-point promise ships with phase 5's API.
+- **`ResultHmap` removal** — phase 3 (plan 22 D6 schedule unchanged). The planner
+  treats it as unplannable-deprecated (D3).
+- **Fixing any of the §2.4 live findings** in the legacy pathways (e.g. the video
+  `reverse` bug, the surface double-filter). Phase 3/4 deletes those code paths; fixing
+  them now would churn the shadow baseline twice. The one exception is documented in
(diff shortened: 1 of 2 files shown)


blooop:
## Review corrections applied (`704719ef`)

All ten findings addressed as doc edits; no design rework. Each was re-verified
independently before editing rather than taken on faith — two by execution.

### The HIGH finding: D6's never-crash mechanism was itself fatal on the reject path

Confirmed by execution. `report_render_failure(what: str, exc: BaseException)` calls
`traceback.format_exception_only(type(exc), exc)` (`render_failure.py:51`); a frozen
`Reject` yields `AttributeError: 'Reject' object has no attribute '__suppress_context__'`.
It would also misbehave in `logger.error(..., exc_info=<Reject>)` and violate D1's own
strict-`ty` mandate.

D6 now specifies **two dispositions instead of one**, and takes the reviewer's design
point that these are different states:

| Outcome | Handler |
|---|---|
| planner raised (a **bug**) | `report_render_failure` — unchanged; "failed to render" is honest |
| planner returned a **`Reject`** (a deliberate decision) | new sibling `report_plan_rejected` + `PlanRejectedWarning`, WARNING-level, own pane wording |

A budget reject ("2,400 tabs exceeds the 20-tab budget; pin `Aggregate(...)`") is expected
behavior, and labelling it "⚠️ failed to render" would teach users to ignore genuine
failure panes. Also withdrew the claim that a `Reject` "is the visible pane without
reformatting" — `reasons`/`suggestions` are tuples and need formatting. Phase 3 owns
implementing the sibling (§10).

### The registry-derivation premise cited the wrong function

Also confirmed by execution: `result_kind()` returns `str` and yields `"unknown"`
off-registry; the typed producer is `result_spec(rv).kind`. The natural coercion
`ResultKind(result_kind(rv))` is **type-clean but raises `ValueError` at runtime** — a
crash at the ingress boundary of a never-crash plan. So plan 24 §2's caveat is *not*
"satisfied by design" as D1 claimed; §2.5 gains a trap table and D4 an explicit
non-coercing disposition (unregistered param → omitted with a log note, never coerced),
plus a new §8 grep banning `result_kind(` inside `bencher/grammar/`. Added the two real
no-silent-link guards I had missed, which strengthen the case:
`test/test_result_missing.py:244` and `bencher/variables/parametrised_sweep.py:96`.

### Withdrawn staleness claims (the ones that mattered most)

These were licence to edit an owner-accepted document, so I re-checked each with
`git log --diff-filter=A`. **A6 was committed `cead5fff`, 2026-07-30 21:18 UTC.**

| Claim | Verdict |
|---|---|
| "a fourth gate layer A6's count predates" | **False.** The `plt_cnt_cfg` gate dates to `0ae9f6ca` (2026-04-26); the `requires`/priority ladder was complete by 07-07. Nothing postdates A6, and A6 counts *shape* dialects — its three is right. |
| "the surface has grown since the audit, in three ways" | **False for 2 of 3.** `tabular_spec` (`ed005ac5`, 07-29), `xy_scatter` (`fb39d9ce`, 07-28) and `rerun_summary` (`11e91171`, 07-30 18:22 UTC) all predate A6. Only `_pane_over_time_dataset` (`48b64798`, 07-31) is genuine post-A6 growth — created by phase 1 itself. Reworded to "A6 did not enumerate them". |
| `RerunSummaryResult` is A6's "second hand-synced rerun recursion" | **Wrong count.** A6's two are both in `rerun_result.py` — `_log_to_rerun` (`:177-300`) and `_build_blueprint_contents._recurse` (`:404-470`). Mine is a **third**, so phase 4's deletion budget is three, not two. Corrected in §2.2 and §3. |

§2.6 is restructured into **three genuinely stale items** (header "no implementation yet";
`hmap_kdims` line drift; finding 21's fourth instance) and **four clarifications** where A6
is correct — including Law 1's `ResultHmap` "removed", which states target architecture, so
flagging it as stale contradicted my own §2.1.1.

### "Byte-identical gallery" was not a checkable claim

`generate-docs` writes only into gitignored trees (`.gitignore:204-206`;
`git ls-files docs/reference/meta/` → 0 files), and `ci` depends on `generate-examples`
(221 tracked files), not `generate-docs`. Promoted the **isolation grep** — nothing under
`bencher/results/` or `bencher/plugins/` imports `bencher.grammar` — to *the* acceptance
artifact: it proves shadow-only structurally (no rendering path can consult the planner,
so nothing it produces can reach a pixel) *and* Law 9 layer 1 (neither hashed module can
see it), and unlike a gallery diff it is actually runnable.

### Policy step order was circular while claiming determinism

Law 7 step 1 needs step 7's mark; step 7 depends on how repeat was handled. I inherited
that verbatim and then pinned determinism three ways, which would have been vacuous. D5
now specifies a **bounded two-pass provisional-mark resolution** with a convergence bound
that is itself a test (a mark table needing a third pass yields `Reject`, never a loop).
Scoped to step 1↔7 per the reviewer — step 4's frame check is kind-derived and not
circular.

### Remaining fixes

- **The one P1 execution blocker:** `SpreadStat` was placed in three modules at once.
  `channels.py` owns it (`KindCaps.spread_stat` needs it, and `plan.py` already imports
  `Channel`, so the alternative is an import cycle). D1 now fixes module ownership and
  the one-way dependency order explicitly.
- Removed the phantom `LayoutPlan` (the type is `SweepPlan`); gave `Pins`/`Prefs` real
  field lists so an agent need not invent them; dropped the spurious
  `warnings.catch_warnings()` note (iterating `RESULT_SPECS` instantiates nothing).
- `rerun_summary`'s `reverse` bug marked **latent, not live** — the only `reverse=True`
  in `bencher/` is `video_summary.py:97` into video's own method.
- **Demoted OD1/OD2/OD4/OD6 to plan decisions** (§6a) — each already carried a
  recommendation with rationale. Only the `ResultVec` deferral (which contradicts an
  owner-accepted Law 7 invariant) and, borderline, plan-at-collect timing remain owner
  calls. §9 reconciled: one substantive deviation + four scope placements, and the PR
  body's count now matches.
- Wrong-precedent citations that would have stranded an implementing agent:
  `test_render.py` has no subprocess/`sys.modules` technique (it walks `gc.get_objects()`
  via `_count_plot_objects()`, `:16-24`) and **no test in the repo** asserts on
  `sys.modules`, so the plan now says so explicitly; `test_explain_selection.py` uses
  list-equality (`:64-68`), not inline expected strings; the golden-hash update procedure
  is the comment block at `test_hash_persistent.py:727-739` with constants at `:740-744`
  and the class spanning `:766-846`; `report_render_failure` is **not** exported from
  `bencher/__init__.py` (`:186` is `RenderFailedWarning`); #1030 is plan 23 P4's registry
  PR, which §2.1.1 had wrong while §2.1.2 had right.
- Range overruns trimmed (`xy_scatter_result.py:56-74`, `PlotMatchesResult:117-178`,
  `test_cartesian_pil_renderer.py:143-180`); `_build_blueprint` corrected to module-level;
  `target_dimension` value **1** added to §2.3's list; the registry ladder given its true
  eight-step order including the easily-missed `plt_cnt_cfg is None` gate.
- README risk column → "Medium" (matches every other row's vocabulary).

### Validation

`TMPDIR=$(mktemp -d) PYTEST_DEBUG_TEMPROOT=$TMPDIR pixi run ci` → **2591 passed, 3
skipped, 0 errors**; `pixi run prek` green including codespell. The private temp root also
confirms the earlier 88 errors were shared-`/tmp` collisions with concurrent agents, not
anything in this branch.

PR #1054, opened 2026-08-03
Enable ty's remaining four Tier-B rules (plan 23 P12b)

blooop:
## Summary

Plan 23 **P12b** — enables the four `ty` Tier-B rules that P12 left. Tier B is now fully enforced.

100 diagnostics cleared. 44 of them traced to two representation defects rather than annotation noise, which is the finding worth carrying:

| Defect | Diagnostics | Fix |
|---|---|---|
| `BenchCfg`'s six `param.List` fields defaulted to `None` | 28 of 48 `not-iterable` | `default=[]` |
| `bencher`'s rerun exports bound only inside `try`/`except` | all 16 `possibly-missing-attribute` | 3 placeholders + 2 dead handlers deleted |

### Three live bugs, each with a regression test verified to fail pre-fix

- **`PluginRegistry.explain()` crashed the report build on the plugin *after* a bad capability.** The handler caught into `exc` — the same name holding the exclude set — and `except … as <name>` unbinds `<name>` on exit, so the next iteration hit `NameError: name 'exc' is not defined`. Out of the one branch whose comment says it never crashes mid-run. The existing test registers a single plugin, so there was no next iteration to fail.
- **`BenchReport.publish()` on an unnamed report** died on `None += …` — on *both* debug settings, since `None += ""` raises too.
- **`TimelineShape`/`StrobeShape` could not be extruded**: both inherited a `_deep_copy_recolored` that walks a `children` list they don't have. Surfaced by making `Shape` abstract, which turned it into a missing implementation — which is what it always was.

### Constructive changes behind the rest

- The Cartesian animation's `Shape` is a `Cell` | `Group` sum type. `Group([])` — constructible before, then dead inside `max()` on an empty sequence — is rejected at construction.
- The two meta-generator config tables are frozen dataclasses instead of `dict[str, int | str | None]`.
- `bencher.sweep()` gained `@overload`s: the return type was always decided by the type of `name`, and the flat union threw that away.
- `get_name()` raises instead of returning the `None` its `-> str` denied could happen.

## Breaking change

**`BenchCfg(input_vars=None)` now raises.** Absence of variables is spelled `[]`. No first-party caller passed `None`. Applies to all six list fields.

## Verification

| Check | Result |
|---|---|
| `pixi run pytest` | 2832 passed, 21 skipped |
| `pixi run test-split` (whole suite through the serialize/render path) | 2836 passed, 21 skipped |
| `pixi run ty` (py313) and `pixi run -e py311 ty` | clean |
| `pixi run lint` (ruff + ty + pylint) | 10.00/10 |
| `prek run --all-files` | all hooks pass |
| Coverage | 91% (gate 85%) |

Claims checked by measurement rather than argument:

- **Cache safety** — `BenchCfg().hash_persistent()` run against the pre-change tree and this one gives `67e6af25…` both times.
- **No visual change** — the six pixel-hash regression tests for the shape renderer are unchanged.
- **No docs drift** — `generate-docs` produces a byte-identical tree; comment parity with `main` verified after the dataclass conversion.
- **The gate bites** — seeded violations produce one diagnostic per rule.

## Review notes

A self-review pass found two defects in this PR's own central claim; both are fixed in `ba2dd68c` and recorded in plan 23 §10 rather than quietly amended.

1. **`possibly-missing-attribute` was never actually enabled.** Three of the four rules are error-by-default, so deleting their `"ignore"` lines turns them on. This one ships *off*. Deleting its line left it disabled while every signal agreed it was armed: `ty` green, key absent exactly like the three that worked, `test_rule_not_globally_ignored` passing on `None`. The probe meant to catch this confirmed the error instead, because it passed `--error` on the command line — measuring `ty`, not the checked-in gate. Now spelled `= "error"`, with `OFF_BY_DEFAULT_IN_TY` requiring that for any such rule and a new probe that reconstructs the repo's own rule table. Re-measured honestly: 16 → 0.

2. **Five of the eight rerun placeholders were unreachable.** Only `utils_rerun` imports `rerun` at module scope; the rest defer it into methods. Verified with a `meta_path` hook. The 16 diagnostics split 9/7 — the 7 were the rule correctly reporting a partiality the *code* asserted and the runtime did not have. Those handlers are deleted in favour of unconditional imports. This also removed a `_MissingExtraMeta` metaclass that was both dead and harmful: raising `ImportError` from `__getattr__` breaks `hasattr` and `getattr(x, n, default)`, which only swallow `AttributeError`.

Smaller review catches: a vacuous export test replaced with an AST check that bites; a `subTest` scoping bug that would have buried a real failure; a comment of mine that named `hv.HoloMap` where `hv.DynamicMap` was meant; eight explanatory comments silently dropped by the dict→dataclass conversion, restored.

Deferred with reasons recorded in plan 23 §10 rather than silently: eleven of the fifteen files touched are not Tier-C clean and so are not strict-listed (counts tabled for plan 26 R9 §3); `Group.direction` is still a stringly-typed three-value tag; `bench_result.py` carries a second, divergent placeholder convention for the same optional dependency.

`pixi.lock` is deliberately not in this PR — it carries pre-existing drift from the dependabot merge on `main`.

## Test plan

- [x] Full suite, normal and split-render pipelines
- [x] `ty` gate on both CI matrix entries
- [x] Each of the three bug-fix tests confirmed to fail on pre-fix code
- [x] Both new gate meta-tests confirmed to fail when their premise is broken
- [x] `generate-docs` output byte-identical
- [x] Cache digest unchanged vs. pre-change tree

## Summary by Sourcery

Enable ty’s remaining Tier-B rules and address the real defects they exposed, including safer representations for bench configuration and rerun-gated exports, plus a small refactor of the Cartesian animation shape model.

Bug Fixes:
- Prevent PluginRegistry.explain() from crashing subsequent plugins when a plugin declares an invalid capability.
- Raise a clear error when publishing a BenchReport without a branch or bench name instead of failing with a TypeError.
- Fix TimelineShape and StrobeShape extrusion so repeat/over-time wrappers no longer crash when deep-copying and recolouring shapes.
- Ensure rerun-dependent exports on the bencher package surface are always defined, with clear ImportError messaging when the optional dependency is missing.
- Avoid crashes and type errors around xarray dimension names, Holoview result overlays, and composable video durations by tightening preconditions and coercions.

Enhancements:
- Replace loosely-typed plot configuration dicts with frozen dataclasses in the meta plot generators for clearer structure and safer defaults.
- Refine the Cartesian animation Shape into explicit Cell and non-empty Group variants, backed by stricter construction and regression tests that pin rendering output.
- Add type-aware overloads and guards (e.g. for sweep(), get_name(), BenchCfg list defaults, and subsampling division selection) so annotations match runtime behaviour and Tier-B rules remain green.
- Strengthen plugin, sweep, and optional-extra tests and planning documentation to codify new invariants and fully enable ty Tier-B checking across strict-listed files.

Documentation:
- Update plans and handover documentation to reflect completion of plan 23 P12b and Tier-B rule enablement, and to record remaining Tier-C cleanup and follow-up ownership.
- Extend the changelog with the newly fixed runtime bugs and the behavioural changes around BenchCfg defaults and rerun-gated exports.

Tests:
- Add regression tests for plugin capability handling, unnamed BenchReport publishing, Shape extrusion wrappers, optional rerun exports, ty gate configuration, and various sweep and blob-store corner cases.
- Update existing renderer, sweep, and config tests to cover the new Shape sum type, BenchCfg list semantics, and stricter handling of optional fields.


blooop:
Automated review did not run: Sourcery is rate-limited again (weekly 500k diff-char cap), same as on #1025/#1026. Its comment above is a generated diff summary, not findings.

Per plan 23's handover §6.5, which says to budget for a manual or subagent adversarial pass when this happens, I ran four independent reviews over the diff before opening this PR — correctness, test quality, type/data-model design against the repo's own constructive-modeling doctrine, and silent-failure hunting.

That pass earned its keep: it found two defects in this PR's own central claim, both fixed in ba2dd68c and written up in plan 23 §10 rather than quietly amended.

1. `possibly-missing-attribute` was never actually enabled. Three of the four Tier-B rules are error-by-default in ty; that one ships **off**. Deleting its `"ignore"` line left it disabled while three independent signals said otherwise — `pixi run ty` green, key absent exactly like the three that worked, and `test_rule_not_globally_ignored` passing on `None`. The seeded probe meant to catch it confirmed the error instead, because it passed `--error` on the command line and so measured ty rather than the checked-in gate.

2. Five of the eight rerun placeholders were unreachable — only `utils_rerun` imports `rerun` at module scope. The 16 diagnostics split 9/7, and the 7 were the rule correctly reporting a partiality the *code* asserted but the runtime did not have.

Both fixes are guarded by tests verified to fail when their premise is broken. Flagging this so a human reviewer knows the automated pass was absent and what stood in for it.

blooop:
Second review round, over the fix commit `ba2dd68c` itself. It found that **three of the four tests that commit added could pass while the property they assert was broken** — fixed in `19569731`, since a test with a false-pass hole is worse than no test.

- `test_gated_imports_bind_on_both_branches` only inspected blocks whose handler was spelled with a bare `ModuleNotFoundError` name. `except ImportError:` — the more common spelling — plus tuple handlers, dotted `builtins.ImportError` and bare `except:` were dropped from the scan entirely. The `assert blocks` guard did not help, because the real `utils_rerun` block keeps the list non-empty. It also harvested handler bindings from *every* handler, so `except ValueError: x = ...` could satisfy a block whose `ModuleNotFoundError` arm was a bare `pass`.
- `test_the_gated_block_is_the_one_module_that_needs_it` matched only literal `rerun*` module names, missing the indirect hoist that breaks `import bencher` identically — a module-scope `from bencher.utils_rerun import x` in one of the three deferred modules. Live, not hypothetical: `variables/results.py` already imports from `utils_rerun`, inside a method.
- `test_repo_rule_table_actually_fires_on_a_seeded_violation` asserted the rule name as a bare substring plus a nonzero exit. ty reports an unknown rule as `warning[unknown-rule]: ... Did you mean \`possibly-missing-attribute\`?` **and exits 1**, so a misspelled — fully disabled — rule satisfied both. This repo has already lived through one ty rule rename (`unused-ignore-comment` → `unused-type-ignore-comment`).

Each fix is verified against the concrete case: passing before, failing after.

Also corrected: a comment I had written in `pyproject.toml` claiming `test_ty_gate.py` "proves each Tier-A/B rule fires against a seeded violation". It does not — two rules have probes, the rest are config reads. That is the same false-claim defect this PR exists to remove, written while removing it. And the `example/` exclusion was justified as "not pulled in by `import bencher`", which is false — `__init__.py` imports `example.benchmark_data`; narrowed to a per-file exemption.

I am stopping the review loop here rather than running a third round. `19569731` is test-hardening plus comment corrections, each verified empirically against the case that motivated it; another round has diminishing returns against the risk of regress. Flagging that as a judgment call rather than leaving it implicit.

PR #1081, opened 2026-08-04
fix: store the blob name in a dataset cell, not its absolute path

blooop:
## The bug

A `ResultDataSet` cell stored the path `materialize_blob` returned, and that path was built from `Path("cachedir").absolute()` — the cache dir's location *in the process that sampled it*.

An `over_time` history outlives that location. It gets tarred on one CI runner and restored at another workspace path, copied between checkouts, downloaded as a tarball for an offline cull, or rendered from a different working directory than the sweep ran in. In all of those, every historical cell pointed at a directory that no longer existed — while the blob it named had been restored intact, under the same content-addressed name.

The payload was reachable. Only the reference was not.

Reproduced by building a 3-event history, relocating the cache dir as `tar -cf` / restore does, and rendering there:

```
                 over_time   rendered runs                     placeholders
before            4          ['run=3']                         3
after             4          ['run=0','run=1','run=2','run=3']  0
```

The placeholder even claimed the payload "could not be loaded and is no longer recoverable from this result", which was false — the bytes were sitting in `blobs/` under the correct hash.

## Why the two halves of the store disagreed

`cache_management` had already worked this out for reachability GC. Its `_blob_name` matched on the **basename**, with a docstring saying why:

> a blob's name *is* its content hash, so a name is a complete and location-independent identity. That keeps references valid when a stored dataset carries an absolute path from a cachedir that has since been moved, copied or renamed — the case where resolving against the current `blobs/` directory would silently declare every reference dead.

So GC was location-independent and rendering was not: one store, two answers to "what identifies a blob". (The GC being right is also why this was never data loss — a relocated cache's blobs are correctly judged reachable, they just could not be read.)

## The fix

- Cells store the blob **name**. `materialize_blob` returns it, so the value a caller gets back *is* the cell value — returning a path is what led to paths being stored.
- New `blob_store.resolve_blob(cell, cache_dir)` turns a cell back into a file, trying the **active cache dir first** and the cell's own literal path as a fallback. Content addressing makes both candidates identical when both exist, so preferring the active one costs nothing and stops a leftover directory (another checkout, a previous CI workspace) from serving reads for the cache actually in use. The literal path still covers a result rendered from a directory with no `cachedir` of its own.
- The name predicate is now public as `blob_store.blob_name`, shared by resolution and by GC instead of defined in both. That also removes `cache_management`'s "must keep matching `blob_store._HASH_CHARS`" coupling comment.

## Compatibility — no `CACHE_VERSION` bump

`resolve_blob` accepts either cell generation, and an absolute path from a 1.118.0 cache is **repaired** by its content name when the recorded directory is gone. Existing histories keep rendering and pick the fix up retroactively, rather than being wiped by a bump.

## Tests

New `test/test_blob_store_portability.py` (19 tests): both cell generations, the resolution order, the failure messages, and the end-to-end CI shape — build an `over_time` history, relocate the cache dir, render in the new location, and require every retained event to render tagged with the run that measured it. That test renders 1 of 4 events on `main` and 4 of 4 here.

Existing blob tests were threaded through the new signature; several stopped needing their own path-to-basename conversion, and `test_blob_store_races._blob_like` drops a `protected-access` disable now that the predicate is public.

Verified: `pixi run pytest` 2855 passed / 21 skipped, `pixi run test-split` (collect/render split forced) 2855 passed, `ty` clean, `pylint` 10.00/10, `prek run -a` clean.

## Not in this PR

Two design questions this turned up, both worth their own discussion:

1. **Blobs are primary storage living inside a directory named `cachedir`.** `clean_orphaned_blobs`' own docstring says a blob is "primary storage, not a recomputable cache" — but downstream treats that tree as disposable (best-effort restore, a 100 GB cull, plans to strip media out of it). The contract of the location and the contract of the contents disagree.
2. **Whether small payloads want a file at all.** A 2-row, 3-column frame serializes to ~2.8 KB of parquet — footer and schema dominate. Below some size threshold the indirection may cost more than it saves, and inlining sidesteps portability, GC and orphan growth entirely.

## Summary by Sourcery

Make blob dataset cells location-independent by storing content-hash names instead of absolute paths and introducing a resolver that loads blobs from the active cache directory, so cached histories remain readable after cache directory moves without a cache version bump.

Bug Fixes:
- Preserve all ResultDataSet history events when the cache directory is moved by resolving blob references via content-hash names rather than stale absolute paths.

Enhancements:
- Introduce a shared blob name predicate and blob resolution helper used by both rendering and GC to keep blob identification consistent and location-independent.
- Clarify result rendering fallback messages to describe blob load failures in terms of attempted locations rather than implying irrecoverable loss.

Documentation:
- Document the new blob cell format and resolution behavior, including cache relocation scenarios, in the blob_store module and changelog.

Tests:
- Add comprehensive blob portability tests covering name-based cells, cache directory relocation scenarios, legacy absolute-path cells, failure messaging, and end-to-end over_time history rendering across moved caches.
- Update existing blob, GC, data model, and result collector tests to use the new blob name format and resolution semantics.


blooop:
Post-merge review (this landed before I finished; filing the notes here and a follow-up issue rather than dropping them).

The diagnosis is right and the core move is right. Cells storing a *location* was the bug; sharing one `blob_name` predicate between GC and the render path fixes a real split-brain where one store gave two answers to "what identifies a blob". The `_BLOB_FORMATS` table is a genuine simplification — it deletes an unreachable-by-construction raise rather than working around it. Tests are strong: both portability axes end to end, resolution ordering, candidate dedupe, and format-table invariants.

What follows is what I'd have pushed on pre-merge.

## Main issues

**1. The composed failure is still broken, and there is no escape hatch.**

`blob_store.py:14-18` names "an offline cull downloads a tarball and works on it wherever it landed" as motivation. That case is *both* axes at once: the active `./cachedir` misses, and the recorded hint points at a dead path. Neither candidate resolves. Meanwhile `render_report()` (`render.py:78`) and the CLI parser (`render.py:151-165`) offer no way to say where the blobs actually are — the resolution order is entirely inferred and entirely unoverridable. An explicit `cache_dir` threaded through `render_report(..., cache_dir=)` plus a `--cachedir` flag (and/or `BENCHER_CACHE_DIR`), passed down to `load_blob`, makes every remaining case recoverable.

**2. With no consumers on >=1.118.0, the legacy generation should be deleted, not supported forever.**

The third resolution candidate carries real weight: the `Path(text) != Path(name)` branch in `_blob_candidates` and its dedupe subtlety, `_parse_or_raise` accepting paths, ~5 tests, and a third branch that can serve a file from a directory nobody asked about. With a `CACHE_VERSION` bump, `resolve_blob` collapses to "name -> active dir, else hint" and `load_blob` takes names only.

`plans/27-cache-version-bump-ledger.md` is a standing ledger of items waiting for exactly this bump, and its own rule is that a bump should land the whole list. The "no bump" section of the PR body is arguing for compatibility with a population of zero, and this is the cheapest moment the project will ever have to pay for it.

(Keep `blob_name`'s basename matching regardless — GC needs to stay robust against a stray path in a walked object, which is a separate justification from reads.)

**3. The hint is stamped on a branch, not at the point blobs are written.**

`bencher.py:905` sits inside `if calculate_results:`. It works today only because a cached `bench_res` was stamped before `cache_results` pickled it — an emergent property that nothing enforces. Any future path producing a `BenchResult` without going through that branch silently loses the hint and reintroduces the foreign-cwd bug with no test failing. Stamping where blobs are actually written (`result_collector._materialize_dataset_value`, which already calls `collect_cache_dir()`) ties the invariant to the fact it depends on.

Related coverage gap: both end-to-end tests set `clear_cache=True` on every run, so nothing covers **a benchmark-cache-hit result still carrying the hint** — precisely the path the placement argument rests on.

## Smaller things

- **`resolve_blob` has no production caller.** `bench_result_base.py:1277` uses `load_blob`, which goes straight to `_locate_blob`. So the module docstring's "`resolve_blob` is the one place that turns a cell back into a file" (`blob_store.py:18`) is not true — `_locate_blob` is. Cheapest fix: have `load_blob` call `resolve_blob`. Otherwise two public entry points duplicate the `cache_dir`/`fallback_cache_dirs` plumbing.
- **`DEFAULT_CACHE_DIR` is introduced but not adopted.** `cache_management.py` still repeats `cachedir: str = "cachedir"` in ten signatures, plus `bench_report.py:315`. Single-source-of-truth is this PR's own stated value in the `_BLOB_FORMATS` refactor; it should reach the cache-dir literal too.
- **`docs/caching.md:211-215` is now stale** and untouched. It is the doc asserting a saved result "still renders in another process" — the exact claim whose mechanism this PR replaces, and which was false for a foreign cwd before it. CHANGELOG and `plans/architecture/A0` were updated; this was not.
- **The `resolve_blob` docstring overstates active-dir-first.** "Content addressing means every candidate holds identical bytes when several exist" is sound *given* no truncation collision, but the digest is truncated to 64 bits and that risk is documented at `blob_store.py:48-59`. Previously a cell resolved within its own cache; now a name from cache A can be served out of cache B. Negligible at the birthday bound, but the docstring should acknowledge the dependency rather than assert identity flatly.
- **The placeholder text got less useful.** `"see the log for the locations tried"` (`bench_result_base.py:1287`) — a rendered HTML report is routinely read without the log. The old message was wrong; this one is correct but unactionable. The blob name, at minimum, belongs in the pane.
- `blob_cache_dir_hints` uses `getattr(dataset, "attrs", {})` to tolerate `None`; an explicit `dataset is None` check says what is meant and does not silently accept arbitrary objects.
- `_blob_candidates`' comment says `Path(text) != Path(name)` detects "the cell is a bare name" — it detects normalized inequality, so `"./abc.parquet"` is dropped as a literal candidate. Harmless, slightly overstated.
- Test nit: `_tmpdir()` at `test_blob_store_portability.py:1445` has a function-local `import tempfile`; module scope is the place for it.
- **PR body drift:** the "The fix" bullets describe only two candidates and never mention `BLOB_CACHE_DIR_ATTR`, which the CHANGELOG and the code both do. That body is now the merge commit message.

Follow-up issue with the actionable subset: #1084

